PR #80, opened 2018-07-26
Automate json doc generation. Remove checked in api.json.

BugDiver:


package.json
```diff
@@ -39,10 +39,10 @@
 		"proxy-from-env": "^1.0.0",
 		"rimraf": "^2.6.1",
 		"recast": "^0.12.8",
-		"repl.history": "^0.1.4"
+		"repl.history": "^0.1.4",
+		"documentation": "^5.5.0"
```
nehashri: The documentation dependency need not be a production dependency. The documentation does not change very much and it needs to be generated only once. It can be generated as a post build/pre-publish task and the output can be published.

Issue #82, opened 2018-07-30
Enable basic authentication


ajoecker:
### Expected behavior
When browsing a webpage, that requires basic authentication, I expect to be able to pass taiko the credentials

### Actual behavior
When calling `goto` with a webpage that requires basic authentication, the script holds and does not proceed.

### Versions
```
$ node -v
v8.11.3

taiko
Version: 0.0.9
```

zabil:
Add a parameter to the `goto` method that allows passing header values to url for e.g

```
goto('http://example.com', {
    headers: {
        Authorization: `Basic ${encodedCredentials}`
    });
```

zabil:
The problem is there are too many header fields
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Standard_request_fields

This will bloat the API and make discovery of the other API's difficult. Also headers maybe repeated across request where it's much easier to declare a variable for re-use

For e.g.
```
var commonHeaders = { headers : ....}
```

This JSON can be merged, overwritten etc.

And visually it's nice to see all the headers fields posted at a glance.

```
goto('http://example.com', {
    headers: {
        Authorization    : `Basic ${encodedCredentials}`,
        From             : 'someone@email.com',
        Proxy-Authorization   : Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==
        User-Agent                : 'Firefox'
    });
```

NivedhaSenthil:
Fix should be available. Authentication can be done by setting extra headers like below example. 

```
(async ()=>{
    await openBrowser({headless:false});
    await goto("https://postman-echo.com/basic-auth",{headers:{'Authorization':'Basic cG9zdG1hbjpwYXNzd29yZA=='}});
    await closeBrowser();
})();
```

PS: Invalid credentials will not be handled with this, as browser allows to do multiple retries on invalid credentials when set on headers.

SanjuSona-19:
I want to keep a login page to be in login even after the chromium is closed. how to achieve this. is there any command is there for keep the cookies not cleared in chromium and to retrieve data from the cookies to avoid multiple login.
help me out to clear this.

SanjuSona-19:
The goto is not waiting until the login process is completing. what to do now? @NivedhaSenthil

BugDiver:
have you tried using a custom profile directory?

Something like `openBrowser({args:["--user-data-dir" : "/some/profile/path"]})` should preserve browser state in the directory mentioned.

[...]

PR #87, opened 2018-08-02
#82 enable basic authentication 

NivedhaSenthil:


lib/networkHandler.js
```diff
@@ -1,20 +1,48 @@
...
+        xhrEvent.emit('xhrEvent', requestPromises[p.requestId].promise);
+    }
+};
+
+const resolveXHR = p => {
+    if (requestPromises && requestPromises[p.requestId])
+        requestPromises[p.requestId].resolveFunc.call();
+};
+
+const authenticate = async (event) => {
```
zabil: Let's not make this an API. Basic authentication requires base64 encoding and is sent across as a header when requesting a url. So it's better to send it as a header parameter in `goto`. Refer comment at https://github.com/getgauge/taiko/issues/82#issuecomment-408880509

NivedhaSenthil: Setting authentication details as part of extra headers does not handle invalid credentials, as we can do multiple attempts to give right credentials. It has to handled separately using [AuthChallengeResponse](https://chromedevtools.github.io/devtools-protocol/tot/Network#type-AuthChallengeResponse)  which requires username and password.

Can it be sent like,

`await goto("https://postman-echo.com/basic-auth",{header: {authentication:['postman','sdg']}}); `

added with a header anyways and used separately or be like how it exists now

`await goto("https://postman-echo.com/basic-auth",authenticate('postman','sdg'));`

IMO putting it in header might be misleading as we dont use it that way and authenticate before hand. Puppeteer handles it the same way with separate [api](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageauthenticatecredentials)

zabil: Does this handle all forms of authentication apart from Basic
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

NivedhaSenthil: Ideally it should handle as we intercept all request with authChallenge to provide credentials but got to test it, looking for ways how to test it.

zabil: Ok, because this is a challenge response. It makes more sense to have an API

NivedhaSenthil: Combination of authentication will not be handled as we are getting only one pair of username and password.

NivedhaSenthil:
Closing this PR as we will be adding auth details as extra headers.

Issue #31, opened 2018-07-06
Not able to interact with elements in new window


saikrishna321:
### Expected behavior
Should be able to interact with elements on new window

### Actual behavior
Not able to find elements on new window

### Steps to reproduce
1. openBrowser()
2. goto('https://getgauge.io')
3. click(text('1194'))

### Versions
```
Version: 0.0.9 (HeadlessChrome/64.0.3264.0)
Node version:  v8.10.0
```

NivedhaSenthil:
Seems like the text "1194" has been changed. I am able to click on "1196" (Github start count when I tried)
Taiko in master is unstable might have bugs, it is being rewritten using Chrome Remote Interface instead of puppeteer.

NivedhaSenthil:
Yes, Taiko should be able to interact with elements in new window.

saikrishna321:
> Taiko should be able to interact with elements in new window

Nope i tried and it didn't work.

NivedhaSenthil:
Can you please give details on what you had tried and what failure occurred ? 
Will help us fix if there is any issue.

saikrishna321:
@NivedhaSenthil sorry for delay response, I missed your comment.

I'm running taiko from the master. How do I provide failure stack trace?

NivedhaSenthil:
Just give us the script you had tried or if you tried from REPL generate script using .code. And what should be the excepted behaviour of the scenario.

saikrishna321:
> goto('https://getgauge.io')
 ✔ Navigated to url "https://gauge.org/index.html"
> openBrowser()
 ✔ Browser and page initialized
> openBrowser()
 ✔ Browser and page initialized
> openBrowser()
 ✔ Browser and page initialized
> goto('https://getgauge.io')
 ✔ Navigated to url "https://gauge.org/index.html"
> click(text('1244'))
 ✔ Clicked element with text "1244"

this would open a new tab with https://github.com/getgauge/gauge and i was trying to interact with elements on the new tab 

> click(text('Sign in'))
 ✘ Error: Element not found, run `.trace` for more info.

NivedhaSenthil:
Was able to reproduce with below script

```
(async () => {
    try {
        await openBrowser();
        await goto('https://getgauge.io');
        await click(link(toLeftOf('Plugins')));
        await highlight(link("Features"));
    } catch (e) {
        console.error(e);
    } finally {
        closeBrowser();
    }
})();
```

[...]

PR #105, opened 2018-08-10
Handle new target

NivedhaSenthil:
- Automatically switches control to new target opened
- Allows to switch between tabs using url or page title

Below scripts demos on accessing elements in newly opened page, switching tab and accessing elements there.

```
(async () => {
    try {
        await openBrowser();
        await goto('https://gauge.org');
        await click(link(toLeftOf('Plugins')));
        await highlight(link("Features"));
        await switchTo('Open Source Test Automation Framework | Gauge'); 
        await click("Plugins");
    } catch (e) {
        console.error(e);
    } finally {
        closeBrowser();
    }
})();
```

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -32,7 +32,7 @@ const connect_to_cri = async (target) => {
                 await Promise.all([network.enable(), page.enable(), dom.enable()]);
                 await networkHandler.setNetwork(network,xhrEvent);
                 await targetHandler.setTarget(criTarget,target,connect_to_cri);
-                pageHandler.setPage(page,xhrEvent,async function () {
+                await pageHandler.setPage(page,xhrEvent,async function () {
                     rootId = null;
                     const { root: { nodeId } } = await dom.getDocument();
                     rootId = nodeId;
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -13,7 +13,7 @@ const ChromiumRevision = require(path.join(helper.projectRoot(), 'package.json')
 const EventEmiter = require('events').EventEmitter;
 const xhrEvent = new EventEmiter();
 const default_timeout = 2000;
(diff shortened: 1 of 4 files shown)


NivedhaSenthil:
This will fix #31

lib/taiko.js
```diff
@@ -64,17 +68,17 @@ module.exports.openBrowser = async (options = { headless: true }) => {
 };
 
 /**
- * Gives the browser version packaged with Taiko.
- *
- * @returns {Promise<string>}
+ * Allows to switch between tabs using URL or page title.
```
Apoorva-GA: An example would help here. Please add an example.

NivedhaSenthil: Updated with examples 👍

lib/taiko.js
```diff
@@ -70,6 +70,10 @@ module.exports.openBrowser = async (options = { headless: true }) => {
 /**
  * Allows to switch between tabs using URL or page title.
  * 
+ * @example
+ * switchTo('https://taiko.gauge.org/') - switch using URL
+ * switchTo('Taiko') - switch using Title
```
saikrishna321: @NivedhaSenthil what if we don't know the URL to switch as it might be dynamic URL?

NivedhaSenthil: We can switch using Title in that case.

saikrishna321: what if two tabs with the same title are present ? how would taiko handle that

NivedhaSenthil: Taiko will pick up the first matching one.

saikrishna321: Is there a way to select the specific window handle currently?

NivedhaSenthil: There is nothing like window handle in taiko for now.

PR #106, opened 2018-08-10
waiting for load event using page lifecycle events

nehashri:


NivedhaSenthil:
Fails for the below script with `Error: Could not find node with given id` as it trying to fetch using the wrong root id. Seems to be not waiting for load to update `rootId` after clicking "About". 

```
const { browser, openBrowser, goto, click,comboBox } = require('taiko');

(async () => {
    try {
        await openBrowser();
        await goto("google.com");
        await click("I'm Feeling Lucky");
        await click("About");
        const inputfield =  await inputField({"placeholder":"Search Doodles"});
        await comboBox(near(inputfield)).select("‪हिन्दी‬");
        await click("Doodles संग्रह");
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
        
    }
})();
```

Issue #102, opened 2018-08-08
Should take z index into consideration when selecting an element


sguptatw:
_From @sguptatw on August 8, 2018 12:32_

**Expected behavior**
Should work with elements of the pop-up currently shown

**Actual behavior**
The element considered is not for the window with the highest z-index

**Version**
```
commit 51de58ab2c042b63be0e5d19a79450de0cd03acd
```

_Copied from original issue: getgauge/gauge#1142_

Apoorva-GA:
As per the discussion with the team,
Taiko will check for a fixed number of elements defaulted to 10, for the match and if there is no match in those number of elements, user will have to provide more appropriate selector in order to fetch the element.

sguptatw:
Another example 
```
const { openBrowser, goto, link, click, highlight } = require('taiko');

(async () => {
    try {
        await openBrowser();
        await goto("google.com");
        await click(link({"title":"Google apps"}));
        await highlight("Gmail");
    } catch (e) {
        console.error(e);
    } finally {
        closeBrowser();
    }
})();
```
> The Gmail highlighted is not of the pop-up

Apoorva-GA:
Fix should be available in commit 888638d275cfebad0747695ebbd1404c60295846

sguptatw:
**Question**

While finding the element, if multiple elements are present. If element(s) is not covered, then should next element should be considered.

**Steps to replicate**
```
const { openBrowser, goto, scrollTo, click, button, link } = require('taiko');

(async () => {
    try {
        await openBrowser({headless:false});
        await goto("http://demos.jquerymobile.com/1.4.5/popup/");
        await scrollTo("sign in");
        await click(link("Sign in"));
        await click(link("Sign in"));
    } catch (e) {
        console.error(e);
    } finally {
        closeBrowser();
    }
})();
```
> Here there are 2 sign in options. One which is covered, the other is not. Which one should be clicked?

sguptatw:
Thanks @Apoorva-GA It works for the above script. However there is another problem

As per the discussion. When there are multiple options for selecting an element, we can narrow down with proximity selectors. I am using the proximity selector `above`. But this is considering an element that is farther away than the one close by -
```
const { openBrowser, goto, link, click, highlight, above } = require('taiko');

(async () => {
    try {
        await openBrowser();
        await goto("google.com");
        await click(link({"title":"Google apps"}));
        await highlight(contains("Gmail"),above("More"))
        await screenshot()
    } catch (e) {
[...]

PR #128, opened 2018-09-05
Handling elements selection when there are multiple elements matching the same condition and some of them are covered by another element #102

Apoorva-GA:


lib/taiko.js
```diff
@@ -244,6 +245,7 @@ const checkIfElementIsCovered = async (elem, x, y, elems, selector) => {
  * @param {string} [options.button='left'] - `left`, `right`, or `middle`.
  * @param {number} [options.clickCount=1] - Number of times to click on the element.
  * @param {number} [options.timeout=5000] -TODO: Timeout value in milliseconds for navigation after click.
+ * @param {number} [options.elemetsToMatch=10] - Number of elements to loop through to match the element with given selector.
```
NivedhaSenthil: Typo elemetsToMatch => elementsToMatch

lib/taiko.js
```diff
@@ -175,6 +175,7 @@ const setOptions = (options, x, y) => {
     options.y = y;
     options.button = options.button || 'left';
     options.clickCount = options.clickCount || 1;
+    options.elemetsToMatch = options.elemetsToMatch || 10;
```
NivedhaSenthil: Typo has to be `elementsToMatch`

lib/taiko.js
```diff
@@ -229,6 +229,7 @@ const checkIfElementIsCovered = async (elem, x, y, elems, selector) => {
         isElemAtPoint = await checkIfChildOfOtherMatches(elems, parseInt(x), parseInt(y));
     if (!isElemAtPoint)
         throw Error(description(selector) + ' is coverred by other element');
```
NivedhaSenthil: Guess this will break throwing error for the first element itself and not continue, has to be done after the loop. Can you please recheck ?

Apoorva-GA: yes, missed this while merging the conflicts.

PR #136, opened 2018-09-06
Check if element is writable before write

nehashri:
- Checks if document has a focused element before writing.
- Checks if element is writable before writing to it.

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -417,25 +417,58 @@ module.exports.write = async (text, into, options = { delay: 10 }) => {
             desc = `Wrote ${text} into the ` + description(selector, true);
             return;
         } else {
-            if (await _getActiveElementTag() !== 'BODY') {
-                await _write(text, options);
-                desc = `Wrote ${text} into the focused element.`;
-                return;
-            }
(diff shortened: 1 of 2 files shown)


lib/taiko.js
```diff
@@ -417,25 +417,58 @@ module.exports.write = async (text, into, options = { delay: 10 }) => {
...
         } else {
-            if (await _getActiveElementTag() !== 'BODY') {
-                await _write(text, options);
-                desc = `Wrote ${text} into the focused element.`;
-                return;
-            }
-            throw new Error('Not focused into a writable field. Please use "write(\'admin\', into(\'Username:\'))" or increase the delay.');
+            await _waitForFocus(options.timeout || default_timeout);
+            await _write(text, options);
+            desc = `Wrote ${text} into the focused element.`;
```
Apoorva-GA: Like we discussed, We should check if taiko actually wrote the text in the focused field and then give the success message.

nehashri: Since we need to assert on all actions to check if they have actually succeeded or not and not just for the write api, I will be creating another PR for the same.

Apoorva-GA: Sure. Lets track it as a separate issue as well. So that @sguptatw can test it better.

Issue #159, opened 2018-09-14
Should be able to navigate to the URL using goto on the instance of browser launched in the current run, not on the previously started browser instance. 


ShilpaGopal:
### Steps to reproduce
1. Make sure you have chromium instance running already
2. openBrowser()
 - this open new browser chromium instance
3. goto("some URL")
 - creates a new tab in the previously running chromium instance instead of the later one.

### Expected behavior
Should be able to create a new tab in the newly launched browser instance unless the user wants to connect to the existing browser instance. 

### Versions
taiko - 0.2.0

NivedhaSenthil:
Fix should be available in a9ea9fa

Issue #110, opened 2018-08-13
Allow user to configure the port


BugDiver:
### Expected behavior
The port should be configurable.

### Actual behavior
A user can not configure the port. Currently, it's hardcoded to `9222`

This might cause an issue if some other process is using the same port.

NivedhaSenthil:
Fix should be available in a9ea9fa

NivedhaSenthil:
eg: `openBrowser({port:9222})`

sguptatw:
Taiko documentation needs to be updated for this.

NivedhaSenthil:
Have updated the docs.

PR #161, opened 2018-09-17
Handle multiple browser instances

NivedhaSenthil:
- make port and host configurable
- use port 0 to connect to any open port instead of hard coding 9222
- use user profile to make browser instances independant


NivedhaSenthil:
This will address #159 and #110

Issue #192, opened 2018-10-01
Need a way to enter combination keys


rajeshvaran:
There should be a way to enter the combination of keys like 'Shift + Tab' for the press function.
Currently, the below sequence of command is not actually doing the intended behavior of shift + tab

Press('Shift')
Press('Tab')

### Expected behavior
A combination of shift + Tab is pressed so that the focus is on the previous element in the sequence.

### Actual behavior
The focus is on the next element in the sequence.

### Steps to reproduce
1.
2.
3.

### Versions
```
Paste `node` and `taiko` versions here.
```

BugDiver:
There are some issues with pressing key combinations. Some key combination works fine other there are a bunch of cases which does not work.
Example - `Ctrl-A`, `Ctrl-R` `Alt-SOME_KEY` etc.

Some of the combinations are os handled actions, not browser handled.
Ref - https://github.com/GoogleChrome/puppeteer/issues/1313

BugDiver:
The decided syntax to press combinations is -
`press(['FisrtKey', 'SecondKey', ....])`
The fix should be availabe in commit  https://github.com/getgauge/taiko/commit/499249fb2a904ee4dc4eb83976d6d1905a3edd07

PR #218, opened 2018-10-12
Added support for pressing key combinations. #192

BugDiver:


lib/taiko.js
```diff
@@ -707,16 +708,25 @@ module.exports.attach = async (filepath, to) => {
...
+};
+
+async function _press(keys, options) {
     await doActionAwaitingNavigation(options, async () => {
-        await inputHandler.down(key, options);
-        if (options && options.delay) await new Promise(f => setTimeout(f, options.delay).unref());
-        await inputHandler.up(key);
+        for (let i = 0; i < keys.length; i++) {
+            await inputHandler.down(keys[i], options);
+            if (options && options.delay) await new Promise(f => setTimeout(f, options.delay).unref());
```
nehashri: The delay need not be in between each key down. It should be after all key down happens and before key up starts.

PR #276, opened 2018-11-27
Generate step from recorded script

BugDiver:


lib/repl.js
```diff
@@ -100,14 +108,38 @@ ${ text }
...
 }
 
+function step(withHooks = false, actions = commands) {
+    if (commands[0].includes('openBrowser()')) actions = actions.slice(0, -1);
+    if (commands[commands.length - 1].includes('closeBrowser()')) actions = actions.slice(1);
+    const actionsString = actions.map(e => {
+        if (!e.endsWith(';')) e += ';';
+        return isTaikoFunc(e) ? '\tawait ' + e : '\t' + e;
+    }).join('\n');
+    const hooks = 'beforeSuite(async () => openBrowser())' +
```
NivedhaSenthil: should be `await openBrowser();` and `await closeBrowser();`

lib/repl.js
```diff
@@ -100,14 +108,38 @@ ${ text }
...
+        '\nafterSuite(async () => closeBrowser());';
+    const cmds = Object.keys(taikoCommands);
+    const importTaiko = cmds.length > 0 ? `const { ${cmds.join(', ')} } = require('taiko');\n\n` : '';
```
NivedhaSenthil: openBrowser and closeBrowser has to be added in imports if not in cmds

lib/repl.js
BugDiver: Not necessarily. It's been invoked as lambda function.
Ref https://github.com/getgauge/gauge-repository/blob/397ac48c9bdd52b35cea57b75773315b97565989/templates/js/tests/step_implementation.js#L9

PR #302, opened 2018-12-21
Fix for near proximity selector

negiDharmendra:
Any element which is within the offset boundary or one of its boundary touches
	  the offset boundary is considered as nearer to the reference element.

           --------------------------------  <---- offset boundry
          |                                |
          |      =================         |
          |      #  Reference     #        |
          |      #  Element       #        |
          |      ==================        |
          |                                |
           --------------------------------


lib/taiko.js
```diff
@@ -1439,16 +1439,27 @@ module.exports.below = selector => {
  *
  * @param {selector|string} selector - Web element selector.
  * @returns {RelativeSearchElement}
+ *
```
NivedhaSenthil: this breaks the docs, try `npm run doc-new`
I feel description can be updated better to convey this. worth mentioning the offset

negiDharmendra: The fix is available in [commit](https://github.com/getgauge/taiko/pull/302/commits/55c404d37f04d771dcc0a0f71d2af7aa388890ae)

Issue #124, opened 2018-08-31
Unable to Automate Popup Notification 


rakeshkmr516:
<img width="372" alt="screen shot 2018-09-01 at 2 13 48 am" src="https://user-images.githubusercontent.com/5263739/44935287-abda3b00-ad8d-11e8-9bba-c698863ac7d7.png">

### Expected behavior
Tell us what should happen
The script should Click button Allow/ Block /X icon of Notification

### Actual behavior
Tell us what happens instead
Unable to perform operations on that Notification 
### Steps to reproduce
1. Navigate to http://google.com
2. Write Text
3. Press Enter

Paste `node` and `taiko` versions here.
Node - v10.8.0
taiko - 0.1.0

Could you please let me know any solution or workaround

rakeshkmr516:
if am using Selenium I have used the below code to disable/block Notifications and Location 
`ChromeOptions opt = new ChromeOptions();
Map<String, Object> prefs = new HashMap<String, Object>();
prefs.put("profile.default_content_setting_values.notifications", 2);
prefs.put("profile.default_content_setting_values.geolocation", 2);
opt.setExperimentalOption("prefs", prefs);
WebDriver driver = new ChromeDriver(opt);`

Can we set something in taiko to customize browser behaviour

NivedhaSenthil:
@rakeshkmr516  For now there is no way we could do this with taiko. Thanks for bringing this up, we will try to handle this scenario in taiko.

NivedhaSenthil:
Fix available in 27c68da

Debashis9012:
The requirement can be handled by using ```overridePermissions() ``` api.
Tested with below version:
``` 27c68da```
For more information: https://taiko-preview.gauge.org/#overridepermissions

PR #345, opened 2019-01-15
override browser permissions

saikrishna321:
Addressed #124


NivedhaSenthil:
I feel the permissions overriden has to be persisted across tabs unless user explicitly resets it, currently on every tab open new session is created hence the bowser context will get lost. 

Can it be passed as an option for `openBrowser` ? Thoughts ?

saikrishna321:
@NivedhaSenthil 

>currently on every tab open new session is created hence the browser context will get lost.

I tried openTab() and it still persists the permissions

screen shot 2019-01-16 at 6 00 25 pm

lib/taiko.js
```diff
@@ -620,6 +678,32 @@ module.exports.focus = async (selector, options = {}) => {
     return { description: 'Focussed on the ' + description(selector, true) };
 };
 
+/**
+ * Override browser permissions
+ * 
+ * @example
+ * overridePermissions('http://maps.google.com',['geolocation']); 
+ *
```
NivedhaSenthil: Adding reference for all available permissions that can be overridden will be helpful

Issue #349, opened 2019-01-17
Cookie Handling


SrinivasanTarget:
Setting and deleting a cookie via APIs

saikrishna321:
@SrinivasanTarget https://github.com/getgauge/taiko/blob/0a6fb22b60e769de56c6327110109d000718cb27/lib/taiko.js#L406 
We can only set one cookie, this approach does not work for setting multiple cookies to the same domain/URL.

SrinivasanTarget:
yes CRI exposes API's for those too. I'm planning to improve even the existing cookie handling once this [PR](https://github.com/getgauge/taiko/pull/354) is in.

nehashri:
Adding a new API to set multiple cookies makes the API complex. In favour of simplicity of the API's, we have decided not to implement a API to set multiple cookies at once.

saikrishna321:
@nehashri Are you saying we can extend the existing API to accept multiple cookies?

nehashri:
@saikrishna321 No, We had a small discussion and we decided that we will not extend the api or create new ones to support setting multiple cookies at once. This is mainly because, we didn't want to complicate the api or introduce new ones.
The user should be able to use the setCookie api to do this.

PR #350, opened 2019-01-18
Delete cookies handling

SrinivasanTarget:
https://github.com/getgauge/taiko/issues/349


lib/taiko.js
```diff
@@ -381,6 +381,38 @@ module.exports.title = async () => {
...
+};
+
+/**
+ * Deletes browser cookies with matching name and url or domain/path pair.
+ *
+ * @example
+ * deleteCookies("CSRFToken", {url: "http://the-internet.herokuapp.com"})
+ *
+ * @param {string} cookieName - Cookie name.
+ * @param {Object} options
```
nehashri: A brief explanation of what options can be passed would be useful here. Also, an example for deleting cookies wrt to a domain can also be added.

SrinivasanTarget: Done

Issue #351, opened 2019-01-19
Feature suggestion: get browser URL


NicholasEF:
### Expected behavior
Would like ability to get a handle on window's current URL, to run assertions against it, or perform other checks.

### Actual behavior
Taiko does't provide a direct handle on this, but other functions do use:
`runtimeHandler.runtimeEvaluate('window.location.toString()');`

Also don't seem to be able to get a handle on this using `evaluate`, which looks to primarily be intended for browser page elements.

### Proposed changes (can submit a PR if preferred)
`taiko.js`
```
/**
 * Returns window's URL.
 * 
 * @example
 * var myURL = getURL();
 * assert(myURL.includes('/sign_in/'));
 *
 * @returns {Promise<String>}
 */
module.exports.getURL = async () => {
    validate();
    const windowURL = await runtimeHandler.runtimeEvaluate('window.location.toString()');
    return(windowURL.result.value);
};
```
```
module.exports.metadata = {
    'Browser actions': ['..., 'getURL'],
```

### Versions
```
`node 6.5.0` and `taiko 0.5.0`
```

NivedhaSenthil:
>Also don't seem to be able to get a handle on this using evaluate 
 
`await evaluate(()=>{return window.location.toString()})` should give that

Will be great if a PR can be created and the api can be `URL` or `currentURL` to keep it consistent with other apis like `title`

NicholasEF:
Ah OK, I will use evaluate in that format for now. 

JD's fork looks great, very thorough. Would love to see that in master soon, or even a new release, along with all the recent optimizations. Thanks!

PR #354, opened 2019-01-21
JD: Added browser URL attribute fetcher to support #351:

jaydeepc:
** User can fetch various browser url attributes including:
    ** protocol
    ** path
    ** host
    ** search
    ** Full Url

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -2262,8 +2262,8 @@ for(const func in module.exports){
 }
 
 module.exports.metadata = {
-    'Browser actions': ['openBrowser', 'closeBrowser', 'client', 'switchTo', 'setViewPort', 'openTab', 'closeTab', 'overridePermissions', 'clearPermissionOverrides', 'setCookie', 'clearBrowserCookies', 'deleteCookies', 'currentURL'],
-    'Page actions': ['goto', 'reload', 'goBack', 'goForward', 'title', 'click', 'doubleClick', 'rightClick', 'dragAndDrop', 'hover', 'focus', 'write', 'clear', 'attach', 'press', 'highlight', 'scrollTo', 'scrollRight', 'scrollLeft', 'scrollUp', 'scrollDown', 'screenshot'],
+    'Browser actions': ['openBrowser', 'closeBrowser', 'client', 'switchTo', 'setViewPort', 'openTab', 'closeTab', 'overridePermissions', 'clearPermissionOverrides', 'setCookie', 'clearBrowserCookies', 'deleteCookies'],
+    'Page actions': ['goto', 'reload', 'goBack', 'goForward', 'title', 'click', 'doubleClick', 'rightClick', 'dragAndDrop', 'hover', 'focus', 'write', 'clear', 'attach', 'press', 'highlight', 'scrollTo', 'scrollRight', 'scrollLeft', 'scrollUp', 'scrollDown', 'screenshot', 'currentURL'],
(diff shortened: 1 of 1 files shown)


lib/taiko.js
```diff
@@ -373,6 +373,38 @@ const _go = async (delta, options = {timeout: 30000}) =>{
...
+ *
+ * @example
+ * browserURL('protocol');
+ * returns http or https
+ *
+ * @returns {Promise<String>}
+ */
+module.exports.browserURL = async (locationAttr = 'toString()') => {
+    validate();
+    if (['protocol', 'host', 'pathname', 'search', 'toString()'].includes(locationAttr)){
```
zabil: I recommend simplifying this and only returning the string.
Once the string is returned the user can use the standard library.
https://developer.mozilla.org/en-US/docs/Web/API/URL to their hearts content.

saikrishna321: recommend: Renaming the function to currentURL makes more meaning when we have multiple tabs open and also aligns with current existing API

jaydeepc: @zabil agree, but I was just thinking 'toString()' returns the complete URL anyway. Now there are lot of cases where we need to assert on protocol or pathname etc and I thought it will give a faster and easier way for QAs to write tests, saving the time to write codes mentioned in https://developer.mozilla.org/en-US/docs/Web/API/URL.

Thoughts?

zabil: From the QA perspective it makes sense. 

But Taiko tries hard to be the API as flat as possible so that it's discoverable and usable from the REPL something a user can see immediately. 

The other thing is compatibility with something like chai https://www.chaijs.com/plugins/chai-url/ which works on strings to provide the assertion.

```
expect(currentURL()).to.contain.path('/foo');
```
Becomes readable out of the box.

Hope this makes sense.

jaydeepc: Ah okay, got it

jaydeepc: Made the changes @zabil and @saikrishna321

saikrishna321: @zabil aren't we tightly coupling with chai for assertions now. What as use if I use expect.js for my assertions. In this case, i can't do  `expect(currentURL()).to.contain.path('/foo');`

Should we just return the Object with matches https://developer.mozilla.org/en-US/docs/Web/API/URL#Properties and let the user decides.

lib/taiko.js
```diff
@@ -2245,7 +2261,7 @@ for(const func in module.exports){
 }
 
 module.exports.metadata = {
-    'Browser actions': ['openBrowser', 'closeBrowser', 'client', 'switchTo', 'setViewPort', 'openTab', 'closeTab', 'overridePermissions', 'clearPermissionOverrides', 'setCookie', 'clearBrowserCookies', 'deleteCookies'],
+    'Browser actions': ['openBrowser', 'closeBrowser', 'client', 'switchTo', 'setViewPort', 'openTab', 'closeTab', 'overridePermissions', 'clearPermissionOverrides', 'setCookie', 'clearBrowserCookies', 'deleteCookies', 'currentURL'],
```
NivedhaSenthil: This is more a Page action as it gives the url of current page. Can this be moved to page actions ?

lib/taiko.js
```diff
@@ -373,6 +373,22 @@ const _go = async (delta, options = {timeout: 30000}) =>{
     await waitForNavigation(options.timeout, promises).catch(handleTimeout(options.timeout));
 };
 
+/**
+ * Returns window's current URL.
+ * @example
+ * openBrowser();
+ * goto("www.abcd.com");
```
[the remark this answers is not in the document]
jaydeepc: Done

lib/taiko.js
SrinivasanTarget: @jaydeepc I think this change is pending.

Issue #272, opened 2018-11-22
Load a script and support to continue recording


sguptatw:
**Proposed behavior**
Load a script and support to continue recording

**Current behavior**
One can only start recording directly from REPL

**Options**
**Option 1**: `taiko movie.js --continue #or --repl` will run the script (not closeBroser) and launch the repl

**Option 2**: In the REPL `taiko --load <previously recorded script>.js` This will run the script (at last it will keep the browser open) and allow recording to continue. The newly recorded code can be added as a different method in <previously recorded script>.js or the user can specify a different file

saikrishna321:
This is a much-needed feature, it's annoying to repeat the same steps in REPL. Can this be prioritized please?

BugDiver:
To load the script and continue to `repl` expose an api `openREPL` from `taiko`. The user can invoke the api at any point in the script and it will start a repl.

Example:
```
const { openBrowser, goto, openREPL, closeBrowser } = require('taiko');
(async () => {
    try {
        await openBrowser();
        await goto('github.com');
       await openREPL();
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
    }
})();
```

BugDiver:
### Draft 2
Instead of exposing API from `taiko` expose it from a submodule, so that the `openREPL`  is not accessible in `repl`. (if will prevent nested `repl` initialization).

Example:
```js
const { openBrowser, goto, closeBrowser } = require('taiko');
const { repl } = require('taiko/recorder');
(async () => {
    try {
        await openBrowser();
        await goto('github.com');
        await repl(); // starts a REPL here
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
    }
})();
```
When the REPL is initilized the further calls are blocked unless repl is exited.

Debashis9012:
As per the draft2 all the requirements has been satisfied. This has been validated and fixed.

PR #367, opened 2019-01-25
Support to load script and continue repl

saikrishna321:
Address  #272  

Need help from team on two things

1. The load script is not synchronized.
2. Once the REPL is loaded the script is executed and new commands are not stored in REPL history

Command: 

`
taiko --load script.js
`

screen shot 2019-01-25 at 2 33 56 pm

### DIFF
--- a/bin/taiko.js
+++ b/bin/taiko.js
@@ -9,7 +9,7 @@ const { removeQuotes, symbols, isTaikoRunner } = require('../lib/util');
 const observeAgrv = ['--observe','--slow-mo','--watch','-o'];
 const argv = process.argv;
 let repl_mode = false;
-const commands = {'-h':printHelpText,'--help':printHelpText,'-v':printVersion,'--version':printVersion};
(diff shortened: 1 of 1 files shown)


bin/taiko.js
```diff
@@ -28,6 +28,12 @@ async function exitOnUnhandledFailures(e){
     }
 }
 
+async function loadScript() {
+    await runFile(process.argv[3]);
```
NivedhaSenthil: this might break when user tries to use `observe` command in a different order, like `taiko -o --load script.js`

saikrishna321: @NivedhaSenthil what does  `observe ` command do? i tried from master and it always looks for .js file

saikrishna321: Current use of `observe` command is `taiko script.js --observe`,  we send --load before the script.js.

If a user wants to use `observe` with load scrip then `taiko --load script.js --observe` this works as expected

NivedhaSenthil: restricting user to follow order for command line args is not very user friendly, we should be considering some CLI framework since the command line args are growing in runner 
like http://yargs.js.org/

PR #368, opened 2019-01-27
Get speed-index and perceptual speed index

saikrishna321:
This PR will help users to measure the page load performance metric that shows you how quickly the contents of a page are visibly populated. The lower the score, the better.

https://developers.google.com/web/tools/lighthouse/audits/speed-index

Command Added:
```
// Start tracing the browser.
startTracing() 

// Stop tracing the browser.
endTracing() 

// Returns the [Speed Index](https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/metrics/speed-index) and [Perceptual Speed Index](https://developers.google.com/web/tools/lighthouse/audits/speed-index) from the page load that happened between the tracing period.
getSpeedIndex() 

// Returns the tracelogs that was captured within the tracing period. You can use this command to store the trace logs on the file system to analyse the trace via Chrome DevTools interface.

getTracingLogs() 
```

Codesnippet:

```
const { openBrowser, startTracing, goto, endTracing, getSpeedIndex, closeBrowser, getTracingLogs } = require('taiko');
const fs = require('fs');
(async () => {
    try {
        await openBrowser();
        await startTracing();
        await goto('https://webdriver.io/blog/');
        await endTracing();
        console.log(await getSpeedIndex());
        fs.writeFileSync('tracelog.json', JSON.stringify(await getTracingLogs()));
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
    }
})();
```


zabil:
@saikrishna321 @NivedhaSenthil @nehashri 
I think it's time to introduce a plugin mechanism for Taiko to introduce features like these. Because of the flat API nature of Taiko the idea is for the core API to have **user** performable actions. 

This was recording in the REPL will be discoverable. However features like this makes sense from the **tester's** perspective. And they can be added after recording the main script.

I recommend something like a `taiko-diagnostics` library for this. Which cab be loaded to expose a different set of API. That way the core and plugins can be independent and easier to evolve.

Issue #329, opened 2019-01-08
Taiko to allow users to evaluate page script


saikrishna321:
Currently, I don't see any way that Taiko can help users to addScript to the current browser session.

The application I test needs local storage to have access_token to use the app else it redirects the users to the login screen which is time killer and its flaky as its thirty party SDK.

Workaround:

`
    client().Page.addScriptToEvaluateOnNewDocument({
      source:
        'console.log(localStorage.setItem("token",JSON.stringify({"access_token":"eyJ0eXAiOiJKV1QiLCJub25jZSI6IkFRQUJBQUFBQUFDNXVuYTBFVUZnVElGOEVsYXh0V2pUT3g0e"})))'
    });
`

CDP API:
https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-addScriptToEvaluateOnNewDocument

BugDiver:
@saikrishna321 This [plugin](https://www.npmjs.com/package/taiko-storage) can be used to interact with the browser's storage.

PR #384, opened 2019-02-04
329 set local storage

jaydeepc:


jaydeepc:
supports #329

NivedhaSenthil:
IMO it is better to expose `addScriptToEvaluateOnNewDocument` separately so that user can add any scripts to be evaluated on new document load, and have apis to support get, set and delete localStorage .. thoughts ?

jaydeepc:
@NivedhaSenthil I thought so. But that was thinking about security consequences. (If there can be any, dint give it too much thought, but I am always a bit picky about giving full control :)). What do you think?

NivedhaSenthil:
Any scripts to be evaluated can be added via `evaluate` api or by accessing CRI apis directly currently.
with `addToLocalStorage` user will expect values given to be added to localStorage in current context which will not happen with the current implementation, it will get added only when there is a new document load and it will try to add for all new document load  so user will have to add it even before `goto` call which may not be what user will expect from `addToLocalStorage`.

We need to give a broader thought for security, currently chromium should be handling major security
concerns like CORS etc and script is run in sandboxed environment  off-process from chromium restricting access to other processes unless user disables it manually.

PR #392, opened 2019-02-08
consider screenorientation for emulateDevice else default to potrait

saikrishna321:


test/functional-tests/tests/pageActions.js
```diff
@@ -60,6 +62,7 @@ step('Assert url host is <hostName>', async function(hostName) {
 });
 
 step('Assert page navigated back', async function() {
+  waitFor(intervalSecs(5));
```
NivedhaSenthil: I feel we should spend time looking at what is actually causing the `timeout` rather than adding waits since adding waits will make tests flaky anyways. We want to avoid using waits in our test suites to capture any scenario where we have to wait for right responses.

I agree there will be scenarios where application might demand waits, but IMO for this scenario we can spend some time understanding whats happening behind page navigations to handle it as in Taiko. Thoughts ?

lib/emulationHandler.js
```diff
@@ -9,12 +9,12 @@ const setViewport = async(options) => {
         throw new Error('No height and width provided');
     options.mobile = options.mobile || false;
     options.deviceScaleFactor = options.deviceScaleFactor || 1;
+    options.screenOrientation = options.screenOrientation || options.isLandscape ? { angle: 90, type: 'landscapePrimary' } : { angle: 0, type: 'portraitPrimary' };
```
NivedhaSenthil: will be great to update documentation with options available

saikrishna321: @NivedhaSenthil Its already linked 

https://github.com/getgauge/taiko/blob/master/lib/taiko.js#L217

NivedhaSenthil: 👍  I was referring to `emulateDevice` docs as the title suggests and just realised it does not take any options

PR #421, opened 2019-02-26
Incorporate commander.js to handle taiko command line args and flags.

negiDharmendra:
```
Usage: taiko [options]
       taiko <file> [options]

Options:
  -v, --version                 output the version number
  -o, --observe                 enables headful mode and runs script with 3000ms delay by default.
          			pass --wait-time option to override the default 3000ms

  --slow-mod                    similar to --observe option

  -w, --wait-time <time in ms>  runs script with provided delay

  --emulate-device <device>     Allows to simulate device viewport. Visit https://github.com/getgauge/taiko/blob/master/lib/device.js for all the available devices

  -h, --help                    output usage information

```

### DIFF
--- a/bin/runFile.js
+++ b/bin/runFile.js
@@ -1,0 +1,41 @@
+const path = require('path');
+const util = require('util');
+const taiko = require('../lib/taiko');
+const { removeQuotes, symbols } = require('../lib/util');
+module.exports = (file, observe, observeTime) => {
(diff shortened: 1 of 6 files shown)


nehashri:
@negiDharmendra The explanation for `--slow-mod` is `similar to --observe option`. We can remove it if they mean the same thing.

Issue #411, opened 2019-02-20
API to simulate touch


ErRei:
Hey there,

this is my first "Issue" in github, so im not sure how to mark this as a feature request.

I am currently writing tests for a mainly mobile used website. It would be nice to have touch events to interact with the website. 
To have some kind of motion like swipe would be awesome. Maybe touchStart and touchEnd with some coordinates?

### Expected behavior
Interact with the website with touch events. 

### Actual behavior
Not implemented right now.

Thank you in advance.

saikrishna321:
>> I am currently writing tests for a mainly mobile used website

Can you please elaborate more on this? Are you emulating a device resolution and then its mobile site?

ErRei:
We are using a webview to deploy the application on mobile devices. But we are testing outside of this webview and emulate device resolution.

My main reason for this feature request is, that we currently swipe list items to delete them. It would be nice if this is testable with taiko. Maybe I've should have wrote in first place...
But all touch-event would be handy.

saikrishna321:
> that we currently swipe list items to delete them 

Is the element loaded in the dom or only on swipe the element loads?


can you see if these API would help https://taiko.gauge.org/#scrollup

ErRei:
It's loaded but hidden. And it shows up on touch/swipe.
I already tried scroll and drag and drop. Both didn't work for me.

NivedhaSenthil:
There is an api to simulate touch in CDP https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchTouchEvent, we could make use of that to simulate touch events like swipe, tap etc.

NivedhaSenthil:
Closing since `tap` api is implemented in Taiko https://docs.taiko.dev/api/tap/

PR #425, opened 2019-02-27
add support to tap with co-ordinates

saikrishna321:
Address https://github.com/getgauge/taiko/issues/411


saikrishna321:
@ErRei can you help test this?

@NivedhaSenthil Don't merge this PR. I want to add logging and handle few more cases, once @ErRei can confirm it works.

NivedhaSenthil:
IMO it will be better if `tap` behaves same like `click` (i.e) take an element selector to tap instead of coordinates as it will be cumbersome for users to calculate it which will keep changing based on the viewport. Can it be like `await tap(link('taiko'))` , thoughts?

saikrishna321:
@NivedhaSenthil 
> await tap(link('taiko'))

I agree, i wanted know if the tap approach works with co-ordinates and i can improve the PR.

saikrishna321:
> IMO it will be better if `tap` behaves same like `click` (i.e) take an element selector to tap instead of coordinates as it will be cumbersome for users to calculate it which will keep changing based on the viewport. Can it be like `await tap(link('taiko'))` , thoughts?

I have fixed this, can you please review ?

NivedhaSenthil:
Handling if element is covered before tapping like how we check before clicking will be great https://github.com/getgauge/taiko/blob/b26eb83db5d815889311376312dee439fd2ebd33/lib/taiko.js#L712
 
All page actions should waitForNavigation by default, check https://github.com/getgauge/taiko/blob/b26eb83db5d815889311376312dee439fd2ebd33/lib/taiko.js#L2002

and element on which action is being done is highlighted by default on headful mode https://github.com/getgauge/taiko/blob/b26eb83db5d815889311376312dee439fd2ebd33/lib/taiko.js#L2058

saikrishna321:
> Handling if element is covered before tapping like how we check before clicking will be great

Done

>  for (let elem of elems) { 
> All page actions should waitForNavigation by default, check
  
Done

>  const doActionAwaitingNavigation = async (options, action) => { 
> and element on which action is being done is highlighted by default on headful mode

   This won't hold good for tap as the touchInput starts and ends on disptach.
https://github.com/getgauge/taiko/pull/425/files#diff-95a7986b58c9812ec261f27da07c8b36R146

>  const highlightElemOnAction = async (elem) => {

Done

lib/taiko.js
```diff
@@ -1522,6 +1522,38 @@ function textField(attrValuePairs, ...args) {
...
+    for (let elem of elems) {
+        isElemAtPoint = false;
+        await scrollTo(elem);
+        isElemAtPoint = await checkIfElementIsCovered(elem, elems, isElemAtPoint);
+        if (isElemAtPoint) {
+            const type = (await evaluate(elem, function getType() { return this.type; })).value;
+            assertType(elem, () => type !== 'file', 'Unsupported operation, use `attach` on file input field');
+            if (headful) await highlightElemOnAction(elem);
+            const { x, y } = await domHandler.boundingBoxCenter(elem);
+            await inputHandler.tap(x,y);
```
NivedhaSenthil: This can done like,
```
await doActionAwaitingNavigation(options,async () => {
   await inputHandler.tap(x,y);
}); 
```

not sure why you feel it won't hold good.

saikrishna321: Fails - TypeError: action is not a function
    at doActionAwaitingNavigation (/Users/saikrisv/git/taiko/lib/taiko.js:2057:11)
    at module.exports.tap (/Users/saikrisv/git/taiko/lib/taiko.js:1545:19)
    at <anonymous>

NivedhaSenthil: I could see options is missed, first arg for `doActionAwaitingNavigation` is options second is the callback function. Please refer other actions for setting options

PR #426, opened 2019-02-28
[BoyScout] - moved helpers methods

saikrishna321:


lib/helper.js
```diff
@@ -98,12 +98,66 @@ const commandlineArgs = () => {
...
+            throw new Error(`waiting failed: timeout ${timeout}ms exceeded`);
+        sleep(intervalTime);
+    }
+};
+
+const waitForNavigation = (timeout, promises = []) => {
+    return new Promise((resolve, reject) => {
+        Promise.all(promises).then(resolve);
+        const timeoutId = setTimeout(() => reject('Timedout'), timeout);
+        timeouts.push(timeoutId);
```
NivedhaSenthil: this timeout has to be cleared on closebrowser https://github.com/getgauge/taiko/blob/ca5933931bab5b273dd558abf92fc68a8c0da9f8/lib/taiko.js#L164

saikrishna321:
@NivedhaSenthil the failure is because of the http://todomvc.com/examples/react/#/ is down. I'm going to move the failing test to http://the-internet.herokuapp.com

lib/helper.js
zabil: Having generic names like `utilities.js`, `common.js` and 'helpers.js` is an anti-pattern as they become a dump yard of functions not related to each other. 

Not in the scope of this PR but at a later point we must consider splitting files according to category or domains.

saikrishna321:
@NivedhaSenthil can you pls review now? Azure is happy 😸

PR #448, opened 2019-03-15
Continue to REPL after executing  given script. #272

BugDiver:


bin/runFile.js
```diff
@@ -29,6 +33,10 @@ module.exports = (file, observe, observeTime) => {
             global[func] = function () {
                 return realFuncs[func].apply(this, arguments);
             };
+        if (continueRepl) global['closeBrowser'] = async () => {
```
NivedhaSenthil: this can be problem if user has closeBrowser in between the scripts, will try to load REPL whenever there is a `closeBrowser`

bin/runFile.js
```diff
@@ -1,23 +1,27 @@
...
+module.exports = async (file, observe, observeTime, continueRepl) => {
     const realFuncs = {};
     for (let func in taiko) {
         realFuncs[func] = taiko[func];
         if (realFuncs[func].constructor.name === 'AsyncFunction')
             global[func] = async function () {
                 let res, args = arguments;
-                if (func === 'openBrowser' && observe) {
+                if (func === 'openBrowser' && (observe || continueRepl)) {
                     if (args['0']) {
```
NivedhaSenthil: do we wanna open browser in headful mode for load always ? if so when user has `args['0']` set and observe is not used it will be `headless: true` (!undefined)

either we can set observe true for load by default or let user decide if it has to be in observe mode by explicitly saying it

BugDiver: > do we wanna open browser in headful mode for load always ?

I would prefer it to be, coz we always launch REPL in headful mode. If run script it headless there won't be a way to bring the browser in foreground.
> either we can set observe true for load by default

This can be done.

bin/runFile.js
BugDiver: Mostly `closeBrowser` is the last action(run in `finally` block). If the user has it in between,  there would be another `openBrowser` if the script has further actions.

NivedhaSenthil: ya even if there is another openBrowser, the problem is loading multiple REPL and thus REPL becomes not useable. 
Try loading below script,
```
const { openBrowser, closeBrowser, goto } = require('taiko');
(async () => {
    try {
        await openBrowser();
        await closeBrowser();
        await openBrowser();
        await goto("google.com");
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
    }
})();
```

Issue #440, opened 2019-03-11
Confirm api doesn't remove listeners after it is used.


nehashri:
confirm api should deregister its listeners after it is used once. 
Consider the following script,
```
(async () => {
    try {
        await openBrowser();
        await goto("http://the-internet.herokuapp.com/javascript_alerts");
        confirm("I am a JS Confirm", async() => await dismiss());
        await click("Click for JS Confirm");
        await text("You clicked: Cancel").exists();
        confirm("I am a JS Confirm", async() => await accept());
        await click("Click for JS Confirm");
        await text("You clicked: Ok").exists();
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
    }
})();
```
The dialog box is dismissed in one scenario, then the same dialog box should be accepted. 
Since we are not deregistering the listeners, this is currently not possible.

*Note: Similar behaviour is seen in alert and prompt apis*

*Dev note: Taiko seems to be registering duplicate listeners for `page.javascriptDialogOpening` api of CDP.*

Debashis9012:
This issue has been verified and found fixed.
Tested with the below version
``` Version: 0.7.0 (Chromium:74.0.3723.0) ```

felipeespitalher:
We'are facing this problem. I have a scenario where two different dialogs may appear, it's depending on data, so, when I add two `confirm("", accept)` one of those aren't used. 

I'm invoking `closeBrowser`, but it seems not removing those listeners 

Is there a way to remove this listener?

phonginx:
+

PR #449, opened 2019-03-15
Fixed issue #440

negiDharmendra:
- Do not invoke a callback if it's already invoked
        when javascriptDialogOpening event is triggered.


negiDharmendra:

PR #451, opened 2019-03-15
Changed exists() to return a boolean value

rselias:
According to the documentation [here](https://taiko.gauge.org/#elementwrapper), exists() is supposed to have been returning a boolean value since 0.4.0, but currently it is returning a JSON object.

It looks like it was changed to return a boolean back in [this commit](https://github.com/getgauge/taiko/commit/49be4d4ea776ea343d23c0abbc6595920f130b58), but since then there's apparently been some reorganization. The call in taiko.js has been moved to helper.js, and the call in repl.js, which was still returning a non-boolean value and I believe was previously returning to taiko.js, is now returning directly to the user.

As far as I can tell, the JSON object is not needed and can simply be changed to a boolean value. Please do correct me if I'm missing something or there's a better way to go about this.

### DIFF
--- a/lib/repl.js
+++ b/lib/repl.js
@@ -192,8 +192,8 @@ function addFunctionToRepl(target, repl, isPlugin = false) {
                     let wrraper = async (interval, timeout) => {
                         let v = await existsFunc(interval, timeout);
                         if (v)
-                            return { description: 'Exists' };
-                        return { description: 'Does not Exist' };
+                            return true;
+                        return false;
                     };
                     res.exists = wrraper;
                 }


rselias:

NivedhaSenthil:
@rselias That wrapper for exists function is created so that REPL will give neat message back. When used as an API it will return boolean.

rselias:
That isn't how it's currently behaving. Right now, the API call to exists() returns the JSON message instead of a boolean.

NivedhaSenthil:
code.js
```
const { openBrowser, goto, link, closeBrowser } = require('taiko');
(async () => {
    try {
        await openBrowser();
        await goto('google.com');
	console.log(await link('Gmail').exists());
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
    }
})();
```
`taiko code.js` prints `true` for me as expected, only on REPL it will print message as `Exists` or `Does not Exist`. Does it behave differently for you ?

rselias:
Ah, I see how it's supposed to be working, thanks. I misunderstood.

What is the reason the wrapper is replacing the boolean value with the description object, instead of logging a neat message and returning the same value? I'm not clear on the value of having the exists() function return different data in the REPL from the API.

NivedhaSenthil:
That is because of the current REPL's design... currently it logs message based on the return object it expects the API to return object with key `description`.

rselias:
Yes, I noticed - I'm more curious about the reason for that design. Would it be problematic to manually log the message while leaving the boolean, in place of using the wrapper to overwrite the return value? Something like:
```
if (res.exists) {
	let logExists = async () => {
		let msg = symbols.pass + 'Does not Exist';
		if (await res.exists()) msg = symbols.pass + 'Exists';
		console.log(removeQuotes(util.inspect(msg, { colors: true }), msg));
	};
	logExists();
}
```

PR #458, opened 2019-03-19
load plugin from local or global node_modules

saikrishna321:
Address https://github.com/saikrishna321/taiko-diagnostics/issues/13#issuecomment-474296651

@BugDiver Pls review

### DIFF
--- a/bin/taiko.js
+++ b/bin/taiko.js
@@ -62,9 +62,13 @@ function getPossibleModulePaths() {
 function loadPlugin(plugin) {
     try {
         let paths = getPossibleModulePaths();
-        let location = paths.find((p) => { return fs.existsSync(path.join(p, 'taiko-' + plugin)); });
+        let location =  paths.map(p => {
+            if(fs.existsSync(path.join(p, plugin))) {
+                return path.join(p, plugin);
+            }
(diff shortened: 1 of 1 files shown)


bin/taiko.js
```diff
@@ -62,9 +62,13 @@ function getPossibleModulePaths() {
 function loadPlugin(plugin) {
     try {
         let paths = getPossibleModulePaths();
-        let location = paths.find((p) => { return fs.existsSync(path.join(p, 'taiko-' + plugin)); });
+        let location =  paths.map(p => {
+            if(fs.existsSync(path.join(p, plugin))) {
+                return path.join(p, plugin);
+            }
+        }).filter(function(p){return p;})[0];
```
nehashri: The filter function here is not really adding value. Is it needed?

saikrishna321: @nehashri Yes, it's needed. There could be a use case where the user has installed taiko-plugin locally or globally then location would look like 

[ '/Users/saikrisv/git/taiko/node_modules/taiko-diagnostics', undefined ]

`filter(function(p){return p;})` returns [ '/Users/saikrisv/git/taiko/node_modules/taiko-diagnostics' ] then [0]

Happy if you suggest a better way of writing this.

PR #468, opened 2019-03-26
Move files that contain static data to a seperate directory. Rename d…

renjithgr:
This commit contains two changes

- Move files `device.js`, `networkConditions.js` and `USKeyboardLayout.js` to a separate directory since they contain only static data.
- Rename `device.js` to `devices.js`.


renjithgr:

Issue #490, opened 2019-04-17
Support for Coffeescript and other languages


tkshnwesper:
I'm developing a [plugin](https://github.com/tkshnwesper/taiko-react) for taiko which is written in Coffeescript.

During development, I am requiring `taiko` as a `devDependency` to write some functional tests. These tests are also written in Coffeescript (and have a `.coffee` extension) and compiled on the fly to Javascript by Mocha (using `coffeescript/register`) prior to running them.

The following line gives the error message that I am receiving:

https://github.com/getgauge/taiko/blob/01997aa88e293953c4fc3d12fa531172a50ee360/bin/taiko.js#L33

I believe it is caused because there is explicit checking of the file extension, and since I am using a Coffeescript file, it is not allowing me to run my code.

https://github.com/getgauge/taiko/blob/01997aa88e293953c4fc3d12fa531172a50ee360/bin/taiko.js#L32

Do you feel this constraint could be made a little more flexible?

**UPDATE**
I previously closed this issue because I somehow got it working, but now I know what I did to get it to work.

If I installed taiko using `npm install taiko` I would get that error. However, if I simply added the git repository to my `package.json` the program would run flawlessly. [That is also what I am currently doing 😄 ](https://github.com/taiko-react/taiko-react-functional-tests/blob/8f638f37b09f9d3af6e9671c1834dd1178e29971/package.json#L33)

I'm not exactly sure what's at play here, because the code that I have referenced above is also present in the latest commit.

Thanks and regards,
Saurabh

tkshnwesper:
Looks like it started working fine without any modifications.

negiDharmendra:
@tkshnwesper The code you have mentioned is only used for taiko CLI.  As you can see this check [here](https://github.com/getgauge/taiko/blob/01997aa88e293953c4fc3d12fa531172a50ee360/bin/taiko.js#L86)

tkshnwesper:
Happens even when I import taiko as a dependency.

<img width="694" alt="image" src="https://user-images.githubusercontent.com/7362366/56919670-d3c99c00-6ade-11e9-8ad8-ade21a84520e.png">

The problem is that this condition evaluates to true:
https://github.com/getgauge/taiko/blob/9e086d8c9bb76b9ce2c55cf0f0520d97e6ead56e/lib/util.js#L25
That's causing it to take the CLI's course of execution.

You could try it out on [this](https://github.com/taiko-react/taiko-react-functional-tests) repo.
[...]

PR #491, opened 2019-04-17
Support for Coffeescript and other languages

tkshnwesper:
This is a follow up on #490 . Made these changes and got taiko to run my functional tests in Coffeescript.

Just added the `.coffee` extension along with `.js` while checking the extension of the file. Though I reckon that Typescript would require `.ts` and many more languages to come in the future would possibly have their own. Perhaps it might be useful to not do such a check at all.

Also, removed the slicing of the filename string since it was cutting off the last three characters of the filename making `xyz.coffee` into `xyz.cof`. I'm guessing this was done to make a filename like `index.js` into just `index`. From what I've seen `index.js` should also work fine in most cases as well :)

Making these changes has worked out well for me. Haven't yet tried making things work together with Gauge yet, but I have my fingers crossed that any problems wouldn't arise.

Cheers! 🎉

### DIFF
--- a/bin/runFile.js
+++ b/bin/runFile.js
@@ -48,5 +48,5 @@ module.exports = async (file, observe, observeTime, continueRepl) => {
         ret.push(__dirname);
         return ret;
     };
-    require(path.resolve(file).slice(0, -3));
+    require(path.resolve(file));
 };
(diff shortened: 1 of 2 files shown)


Shuunen:
@tkshnwesper @gaugeci why this MR is closed ?

tkshnwesper:

tkshnwesper:
Might have to open a new MR since I deleted my fork of the repo after raising this

PR #506, opened 2019-04-24
Unit test for closerBrowser and textBox

Manikkumar1988:


test/unit-tests/openBrowser.test.js
```diff
@@ -11,6 +12,23 @@ describe(' opens browser successfully',()=>{
...
+        return openBrowser().then(() => {
+            expect(client).not.toBeNull();
+        });
+    });
+
+
+    test('openBrowser should initiate the CRI client object with Browser domain',  ()=>{
+
+        return openBrowser().then(() => {
+            expect(client.Browser).not.toBeNull();
```
NivedhaSenthil: `client.Browser` is CRI's internal domain.. if client is initialised it should be there always we don't have to verify this.

test/unit-tests/textBox.test.js
```diff
@@ -0,0 +1,32 @@
...
+            await goto(filePath);
+        }, 10000);
+
+        afterAll(() => {
+            removeFile(filePath);
+        });
+
+        afterEach(async() => await closeBrowser());
+
+        test('test exists()', () => {
```
NivedhaSenthil: tests can be added for the following cases: 

- input with type text and has label with for, inner element of label and using attribute/value pair  (exists, value)
(e.g) 
```
<label for="username">Username</label><input type="text" id="name" name="name"/>
<label>Do you like peas?
  <input type="checkbox" name="peas">
</label>
```
- textarea and has label with for, inner element of label and using attribute/value pair (exists, value)

PR #509, opened 2019-04-25
483 Making runner's console log event based 

Apoorva-GA:
- Making runner's console log event based 
- Removed the wrapper for `exists` in REPL
- RunFile also uses the descriptions from emitted event.
- REPL now prints the return value along with the description.

### DIFF
--- a/bin/runFile.js
+++ b/bin/runFile.js
@@ -24,12 +24,6 @@ module.exports = async (file, observe, observeTime, continueRepl) => {
                 }
 
                 res = await realFuncs[func].apply(this, args);
-                var eventEmitter = taiko.emitter;
-                eventEmitter.on('success', (desc) => {
-                    desc = symbols.pass + desc;
-                    desc = removeQuotes(util.inspect(desc, { colors: true }), desc);
-                    console.log(desc);
-                });
                 return res;
             };
         else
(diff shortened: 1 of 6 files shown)


bin/runFile.js
```diff
@@ -24,10 +24,12 @@ module.exports = async (file, observe, observeTime, continueRepl) => {
                 }
 
                 res = await realFuncs[func].apply(this, args);
-                if (res.description) {
-                    res.description = symbols.pass + res.description;
-                    console.log(removeQuotes(util.inspect(res.description, { colors: true }), res.description));
-                }
+                var eventEmitter = taiko.emitter;
+                eventEmitter.on('success', (desc) => {
```
NivedhaSenthil: This will add listener many times for all taiko funcs, listener can be added before looping.

lib/taiko.js
```diff
@@ -31,6 +31,8 @@ let chromeProcess, temporaryUserDataDir, page, network, runtime, input, client,
 
 let plugins = new Map();
 
+module.exports.emitter = xhrEvent;
```
NivedhaSenthil: I feel we can use different emitter for descriptions, so that user wont accidentally change the behaviour of other events

test/unit-tests/closeBrowser.test.js
```diff
@@ -9,10 +9,13 @@ describe('close browser successfully',()=>{
     },10000);
     
 
-    test('closeBrowser should return \'Browser Opened\' message',  ()=>{
+    test('closeBrowser should return \'Browser Opened\' message', async ()=>{
 
-        return closeBrowser().then(data => {
-            expect(data).toEqual({ 'description': 'Browser closed' });
+        emitter.on('success', (desc) => {
+            expect(desc).toEqual('Browser opened');
```
NivedhaSenthil: guess assertion is not waiting for listener here, else it should have failed for `Browser opened` which is expected on `closeBrowser`

PR #540, opened 2019-05-10
Fixed xpath for input field

shubhamsc:


test/unit-tests/inputField.test.js
```diff
@@ -0,0 +1,115 @@
...
+        beforeEach(async () => {
+            await goto(filePath);
+        });
+
+        afterAll(() => {
+            removeFile(filePath);
+        });
+
+        test('test exists()', async () => {
+            await expect(inputField({ id: 'user[login]' }).exists()).resolves.toBeTruthy();
```
NivedhaSenthil: this test can be updated to use wrapped in label instead of using attr and value pair

test/unit-tests/inputField.test.js
```diff
@@ -0,0 +1,115 @@
...
+
+    beforeAll(async () => {
+        await openBrowser(openBrowserArgs);
+    }, 30000);
+
+    afterAll(async () => {
+        await closeBrowser();
+    }, 30000);
+
+    describe('with inline text', () => {
```
NivedhaSenthil: `describe` can be `attribute and value pair` since there is no `inline text` here

PR #544, opened 2019-05-13
Fixed for match case sensitive text

shubhamsc:


lib/taiko.js
```diff
@@ -1881,7 +1881,7 @@ function match(text, ...args) {
         let nbspChar = String.fromCharCode(160);
         let textToTranslate = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ' + nbspChar;
         let translateTo = 'abcdefghijklmnopqrstuvwxyz' + ' ';
-        let xpathText = `translate(normalize-space(${xpath(text)}), "${textToTranslate}", "${translateTo}")`;
+        let xpathText = `translate(normalize-space(${xpath(text).toLowerCase()}), "${textToTranslate}", "${translateTo}")`;
```
NivedhaSenthil: translate should actually have taken care of converting to lowercase..seems like this is happening because of the non-alphabetic character in the DOM which is not getting converted to lowercase by translate as it considers only alphabets currently. We may have to add more characters like, 
`translate(text(), 'ABCDEFGHIJKLMNOPQRSTUVWXYZÄÖÜÉÈÊÀÁÂÒÓÔÙÚÛÇÅÏÕÑŒ', 'abcdefghijklmnopqrstuvwxyzäöüéèêàáâòóôùúûçåïõñœ')`  since chrome supports only xpath1.0 or find a way to do it using javascript

NivedhaSenthil:
Closing this as there is no cleaner way to convert all non-alphabetic characters in xpath1.0

PR #576, opened 2019-05-29
Fix for finding text node

shubhamsc:


lib/taiko.js
```diff
@@ -1909,30 +1909,55 @@ function match(text, ...args) {
...
+        if (!elements || !elements.length) elements = await matchContainsElements(e, textToTranslate, translateTo, xpathText);
         return await handleRelativeSearch(elements, args);
     };
     return { get: getIfExists(get), exists: exists(getIfExists(get), descEvent), description: `Element matching text "${text}"`, text: selectorText(get) };
 }
 
-async function matchTextNode(textToTranslate, translateTo, xpathText) {
-    const elements = await $$xpath(`//text()[translate(normalize-space(string()), "${textToTranslate}", "${translateTo}")=${xpathText}]`);
-    if ((!elements || !elements.length)) return await $$xpath(`//text()[contains(translate(normalize-space(string()), "${textToTranslate}", "${translateTo}")=${xpathText})]`);
+async function matchExactElements(e, textToTranslate, translateTo, xpathText) {
```
NivedhaSenthil: can these methods be moved into a separate file as they are well isolated now ? something like textSearch.js

test/unit-tests/textMatch.test.js
```diff
@@ -1,11 +1,20 @@
 let { openBrowser, goto, closeBrowser, text, textBox, toRightOf, evaluate } = require('../../lib/taiko');
 let { createHtml, removeFile, openBrowserArgs } = require('./test-util');
+let test_name = 'textMatch';
 
-describe('text match', () => {
-
+describe('match', () => {
```
NivedhaSenthil: updating test to do goto only once reduces a lot of test runtime since goto adds 1s for every testcase .. will be great to have it the same way

Issue #562, opened 2019-05-22
Load plugins dynamically 


NivedhaSenthil:
## Taiko will be dynamically loading plugins and will let plugins override apis.

- run with env variable like 
```
TAIKO_PLUGIN='android'
TAIKO_PLUGIN='awslambda'
```
- or using taiko runner like `taiko code.js --plugin 'android'` 
- as a fallback can load plugins from dependencies
- export plugin object instead of exporting apis from plugin (eg: `taiko-screencast` plugin apis will exports via object named `screencast` and apis can be accessed like `screencase.start()` ) 
- use convention `taiko-<plugin>` - all taiko plugin should prefix `taiko-`

## Event driven pattern on actions 

- Plugins should not have taiko as dependencies it should use the same used in users code
- `loadPlugin()` api should give `taiko` and `eventHandler` to plugin via callBack - callback is currently named `clientHandler` could be called `init` which will initialize the plugin (may not need an plugin id if dynamic loading is done)
- all other communication between plugin and taiko should be based on event (eg: whenever a connection is done taiko emits an emit so that plugin can listen and update its cri client)

Debashis9012:
This enhancement has been verified and fixed based on the below observation.
Observation:
taiko allowing the plugins to override the apis. 
Commit version: ``` 7493c6afa8565c8482f8f01ada43c32dd1185627 ```

PR #577, opened 2019-05-30
Autoload plugin #562

luciferankon:


bin/taiko.js
```diff
@@ -141,11 +147,12 @@ if (isTaikoRunner(process.argv[1])) {
                         });
                     });
                 } else {
-                    runFile(fileName, observe, program.waitTime);
+                    runFile(taiko, fileName, observe, program.waitTime);
                 }
             } else {
+                loadPlugins(program.plugin);
```
NivedhaSenthil: loading plugins may not be required as `--plugins` will add them to env variables and then `require('taiko')` takes care of loading them

lib/taiko.js
```diff
@@ -2220,20 +2221,41 @@ module.exports.dismiss = async () => {
...
         console.trace(error);
     }
 };
 
+module.exports.loadPlugin = loadPlugin;
+
+
+const overriddenAPIs = {};
+getPlugins().forEach( (pluginName) => {
+    let plugin = require(path.resolve(`node_modules/${pluginName}`));
```
NivedhaSenthil: global paths should also be considered as `TAIKO_PLUGIN='android' node code.js` will break if there is no local node_modules and when done loadPlugins in cli can only set env variables

lib/taiko.js
```diff
@@ -2220,20 +2221,41 @@ module.exports.dismiss = async () => {
...
+    loadPlugin(plugin.ID, plugin.init);
+    module.exports[plugin.ID] = plugin;
```
NivedhaSenthil: plugins need not have IDs as taiko plugins will follow the convention of having name as `taiko-<pluginName>` pluginName has to be unique thus the same can be used to avoid conflicts between pluginIds and the exported object should have key as `pluginName` if different Id is set that convention will break

PR #586, opened 2019-06-01
Fixed incorrect parameters to runFile

SamLeatherdale:
The --load option was not working due to the fact that the parameters used are incorrect. Updated parameters passed to runFile to match signature.


SamLeatherdale:
OK, have now signed Contributor License Agreement.

PR #588, opened 2019-06-03
Added --emulate-network option to CLI options

negiDharmendra:


NivedhaSenthil:
I feel it is better to do it same as `emulateDevice` using env variables to have consistency

PR #593, opened 2019-06-04
Extend waitFor API

shubhamsc:


lib/taiko.js
```diff
@@ -2280,11 +2282,26 @@ module.exports.into = e => e;
  * This function is used to wait for number of secs given.
  *
  * @example
- * waitFor(intervalSecs(5))
+ * waitFor(toMilliSeconds(5))
+ * waitFor("1 item in cart")
```
NivedhaSenthil: update docs for default time to wait for element if not given and param can be `@param {selector|string}`

PR #602, opened 2019-06-07
Use keyDown and keyUp to write

shubhamsc:


lib/taiko.js
```diff
@@ -1013,11 +1013,12 @@ const _write = async (text, activeElement, options) => {
         throw new Error('Element focused is not writable');
     if (headful) await highlightElemOnAction(activeElement.nodeId);
     for (const char of text) {
+        await input.dispatchKeyEvent({ type: 'keyDown', text: char });
         await new Promise(resolve => {
```
NivedhaSenthil: delay in options is meant to add delays between each character press, can be moved after 'keyUp'

Issue #621, opened 2019-06-14
switchTo fails while using url


Debashis9012:
### Expected behavior
Should be switched to specified url

### Actual behavior
Throwing below error message:
``` ✘ Error: No target with given URL/Title found., run `.trace` for more info.```

### Steps to reproduce
```
openBrowser()
 ✔ Browser opened
> goto("http://www.way2automation.com/protractor-angularjs-practice-website.html")
 ✔ Navigated to URL http://www.way2automation.com/protractor-angularjs-practice-website.html
> click(image(toRightOf("Checkboxes")))
 ✔ Clicked image To right of Checkboxes
> switchTo("http://www.way2automation.com/protractor-angularjs-practice-website.html")
 ✘ Error: No target with given URL/Title found., run `.trace` for more info.

``` 
Version: 0.8.0 (Chromium: 76.0.3803.0) 2a3f39e971d0a455720d2e0462975736a99e4b33
```

Debashis9012:
This issue has been verified and found fixed with the version ```e61ef4a016d5ce9997869d7aadf57cfebd63c73b```

PR #626, opened 2019-06-17
fix to switchTo is not working with url #621

luciferankon:


lib/targetHandler.js
```diff
@@ -41,7 +41,7 @@ const getCriTargets = async function (targetUrl) {
         response = { matching: [], others: [] };
         for (const target of targets) {
             target.url = handleUrlRedirection(target.url);
-            if (target.title === targetUrl || target.url.split('://')[1] === (targetUrl.split('://')[1] || targetUrl)) {
+            if (target.title === targetUrl || target.url.split('://')[1] === (targetUrl.split('://')[1] || targetUrl) || target.url.split('://')[1] === (targetUrl.split('://www.')[1] || targetUrl)) {
```
nehashri: There should be a better way to do this. We can compare against the domain, path of the url to achieve this.

PR #629, opened 2019-06-17
Do not initialize plugins subcommand for taiko REPL

negiDharmendra:


bin/taiko.js
```diff
@@ -56,16 +57,21 @@ function setPluginNameInEnv(pluginName) {
...
         console.log(`Invalid value ${networkType} for --emulate-network`);
         console.log(`Available options: ${NETWORK_TYPES.join(', ')}`);
         process.exit(1);
     }
     process.env.TAIKO_EMULATE_NETWORK = networkType;
 }
 
-if (isTaikoRunner(process.argv[1])) {
+function isReplCommand(args) {
+    return args.filter(arg => !(arg.match(/^-/) && !arg.match(/^(-h|--help)/))).length < 3;
```
shubhamsc: In this case, we can open taiko repl with `taiko` only, but we can't open taiko repl with any plugin.
like `taiko --plugin taiko-screencast`. There should be a generic way to launch taiko repl with plugin also.

negiDharmendra: @shubhamsc  Thanks for pointing it out. Please take a look at the latest changes.

PR #632, opened 2019-06-18
remove unused <li> on the doc

CzarScar:
I am not sure this "<li>" is necessary. If the "<li>"needed, please reject this pull request.

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -1644,7 +1644,7 @@ module.exports.link = (attrValuePairs, ...args) => {
 };
 
 /**
- * This {@link selector} lets you identify a list item (HTML <li> element) on a web page with label or attribute and value pairs and proximity selectors.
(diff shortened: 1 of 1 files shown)


lib/taiko.js
```diff
@@ -1644,7 +1644,7 @@ module.exports.link = (attrValuePairs, ...args) => {
 };
 
 /**
- * This {@link selector} lets you identify a list item (HTML <li> element) on a web page with label or attribute and value pairs and proximity selectors.
+ * This {@link selector} lets you identify a list item (HTML element) on a web page with label or attribute and value pairs and proximity selectors.
```
Apoorva-GA: This is to specify that the, `listItem` api identifies the html `<li>` tag and that is required.

Apoorva-GA: With commit : 51102a49067dd7e7b9af671f071b45c5d435db4e docs should now display `<li>` tag properly.
Closing this for now. Thanks for the PR!

CzarScar: Thank you for your quick reply.

PR #665, opened 2019-07-03
Added options to select hidden element to fileField API

negiDharmendra:


lib/taiko.js
```diff
@@ -1767,12 +1768,19 @@ module.exports.inputField = (attrValuePairs, ...args) => {
  */
 module.exports.fileField = fileField;
 
-function fileField(attrValuePairs, ...args) {
+function fileField(attrValuePairs, options = {}, ...args) {
     validate();
+    if( options instanceof RelativeSearchElement) {
```
NivedhaSenthil: this check can be moved inside `getValues` which is used in common for most selectors

Issue #658, opened 2019-06-27
click api does not work with argument "clickCount"


Debashis9012:
### Expected behavior
`click` api should work properly with clickCount argument

### Actual behavior
Click operation did not perform as per the number of click count.

### Steps to reproduce
Kindly follow the below steps in repl:
```
  await openBrowser();
        await goto("https://www.airbnb.co.in/");
        await write("Goa",into(textBox(below("WHERE"))));
        await press("Enter");
        await click(textBox(below("CHECK-in")));
        await click("OK");
        await click(textBox(below("CHECK-in")));
        await click($("//td[@class='_16zigr23'][contains(text(),'27')]"));
        await click($("//td[@class='_16zigr23'][contains(text(),'30')]"));
        await click(button(below("GUESTS")));
        await click($("//div[@class='_9cfq872']//div//div[1]//div[1]//div[1]//div[1]//div[2]//div[1]//div[3]//button[1]"),{clickCount:5}); // Here click api didn't perform for 5 times
       
```

### Versions
```
Commit: 765f95ad2da5de18448caa8890f8715a725ff446
```

Debashis9012:
This issue has been verified and found fixed.
Tested version: ```ead0ab9b52c68989c32abe6ca5e675792bfeb8ae```

PR #666, opened 2019-07-03
handled click count to click specified number of times #658

luciferankon:

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -858,7 +858,10 @@ async function click(selector, options = {}, ...args) {
         options = {};
     }
     if (isSelector(selector) || isString(selector) || Number.isInteger(selector)) {
-        await _click(selector, options, ...args);
+        const clickCount = options.clickCount || 1;
+        for (let count = 0; count < clickCount; count++) {
+            await _click(selector, options, ...args);
+        }
         descEvent.emit('success', 'Clicked ' + description(selector, true));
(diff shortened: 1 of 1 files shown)


lib/taiko.js
```diff
@@ -858,7 +858,10 @@ async function click(selector, options = {}, ...args) {
         options = {};
     }
     if (isSelector(selector) || isString(selector) || Number.isInteger(selector)) {
-        await _click(selector, options, ...args);
+        const clickCount = options.clickCount || 1;
```
NivedhaSenthil: doing `_click` so many times will be costlier as getting element, proximity search, finding if it covered, scrolling element into view  are all done in it.. guess it will be enough if `waitForMouseActions` is done as many times as click count. Can have a different name to differentiate between `doubleClick` and `clickCount` option

lib/taiko.js
```diff
@@ -825,7 +825,9 @@ async function _click(selector, options, ...args) {
     if (!isElemAtPoint)
         throw Error(description(selector) + ' is covered by other element');
     options.x = X, options.y = Y;
-    await waitForMouseActions(options);
+    for (let count = 0; count < options.noOfClicks; count++) {
```
NivedhaSenthil: guess this will also set clickCount option for input events, hence each click will be triggered with so many click conts... looks like clear test is failing for the same

PR #667, opened 2019-07-03
select using index and value from dropdown

luciferankon:


lib/taiko.js
```diff
@@ -1937,6 +1937,9 @@ function dropDown(attrValuePairs, ...args) {
         exists: exists(getIfExists(get), descEvent),
         description: desc(selector, 'label', 'Dropdown'),
         select: async (value) => {
+            if(!isString(value)){
```
NivedhaSenthil: this may not be needed

lib/taiko.js
```diff
@@ -1945,7 +1948,7 @@ function dropDown(attrValuePairs, ...args) {
             function selectBox(value) {
                 let found_value = false;
                 for (var i = 0; i < this.options.length; i++) {
-                    if (this.options[i].text === value || this.options[i].value === value) {
+                    if (this.options[i].text === value || this.options[i].value === value || this.options[i].index === value) {
```
NivedhaSenthil: when an index is given we dont have to iterate that many times, index can be set directly

lib/taiko.js
```diff
@@ -1937,13 +1937,19 @@ function dropDown(attrValuePairs, ...args) {
...
         description: desc(selector, 'label', 'Dropdown'),
         select: async (value) => {
-
             const nodeId = (await getIfExists(get)())[0];
             if (!nodeId) throw `${desc(selector, 'label', 'DropDown')} not found`;
             if (headful) await highlightElemOnAction(nodeId);
 
             function selectBox(value) {
                 let found_value = false;
+                if(this.options.length > value.index){
```
saikrishna321: Why value.index and not just value which would simplify the API

From:
```
await dropDown(below('Reason')).select({index:1});
```

To:
```
await dropDown(below('Reason')).select(1);
```

luciferankon: @saikrishna321 Then there is a possibility that it may collide with other element's value... Meaning consider two elements, one element with index 1 and the other with value 1. Which one do you select then?

PR #669, opened 2019-07-05
Do not use deprecated API in docs examples

chris-kobrzak:
The documentation mentions `textField` and when you run this example, you get the following warning in the console:

    DEPRECATION WARNING: textField is deprecated. Use textBox to select inputField with type text and textarea instead.

So this has been swapped with `textBox`.

Apologies for a number of changes in the diff - it's my editor that automatically strips trailing white space. Probably worth disabling white space changes when reviewing this PR.


NivedhaSenthil:
Thanks for the PR 👍

Issue #352, opened 2019-01-21
Refractor and split taiko.js


SrinivasanTarget:
The idea behind this card is to split taiko.js (2.5k lines) into separate files/classes,

- Browser specific functions
- Page specific functions
- Element specific., etc

zabil:
(Something to keep in mind while splitting)

It's originally kept as a single file to make sure that user has
* Only one import
* Jump to documentation and source in one step from the IDE
* Easy to generate documentation

PR #676, opened 2019-07-10
#352 elements pull out

NivedhaSenthil:
- pull out element fetching logic into `elementSearch.js`
- pull out elementWrapper helps into `elementWrapperUtils.js`
- introduce eventbus to share events and avoid coupling of domains
- make all handlers get updated domain using event handler rather than setting from `connectToCri`
- fix tests for above changes


NivedhaSenthil:
since `createdSession` is a valid event that needs more than 10 listeners as all handlers 
have to listen to it updated the max listener count to 20 leaving a buffer for plugins to listen. Currently there is no way to add max listener count for particular event hence set it on the emitter.

PR #687, opened 2019-07-14
added pre-push hook

saikrishna321:
Run unit test as a pre-push hook to make sure things work before CI gets unhappy 😒

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,11 +17,6 @@
 		"install": "node lib/install.js",
 		"update-json-doc": "node build/updateJsonDoc.js"
 	},
-	"pre-commit": [
-		"test:api",
-		"lint",
-		"update-json-doc"
-	],
 	"repository": {
 		"type": "git",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -36,6 +36,11 @@
 		"chromium_revision": "662092",
 		"chromium_version": "76.0.3803.0"
 	},
+	"husky": {
```
BugDiver: Since now we are using `husky` to manage hooks, can you update the `pre-commit` hook as well?
Then we can remove the `pre-commit` dependency as well.

saikrishna321: Done!

saikrishna321:
@BugDiver Is there anything else you think I need to address to get this merged?

BugDiver:
@saikrishna321 Nothing which I can think of. Somehow it lost track of this.
But I see the builds are broken.

Issue #690, opened 2019-07-15
Add a option to disable step description


geekeren:
### Expected behavior
When I run taiko with option "-s" (--slient), I want No description message in the console logs, like "Browser opened" 、“Navigated to URL“....

```
 ✔ Browser opened
 ✔ Navigated to URL

My console log
My console log
My console log

```
it should be very clean, with only My customized console logs.
```
My console log
My console log
My console log
```

### Context: Why I need that
I want to use taiko as a command to get an access token, I want to call it in shell script and export the taiko output as an environment
```
export USER_AUTH_TOKEN=$(taiko get-acccess-token.js)
```
Then I use USER_AUTH_TOKEN everywhere.

zabil:
You can do this by running the script using node.

For e.g. instead of running
```
$ taiko script.js
```

use

```
$ node script.js
```

geekeren:
> You can do this by running the script using node.
> 
> For e.g. instead of running
> 
> ```
> $ taiko script.js
> ```
> 
> use
> 
> ```
> $ node script.js
> ```

Thanks. But, I have tried, but I should create a project for that, and add taiko as a npm library.
 I intend to use as a command.
Anyway, I think it's better to have a option like "-s silent" for me😀
taiko -s xxx.js

Apoorva-GA:
Fixed as part of https://github.com/getgauge/taiko/pull/692.

Debashis9012:
Above changes has been verified and found fixed. 
Tested version: `Version: 1.0.1 (Chromium: 76.0.3803.0) 5ada00f72c218606b9681f18966987ac606bcaba`

PR #691, opened 2019-07-15
Wrap console log && support --no-log

rsun-thoughtworks:
Fix issue #690 

Provide an option --no-log to disable all log from taiko itself, but still, allow user's script to use console.log

```sh
# will disable logs from within taiko itself
$ taiko user.js --no-log
```


rsun-thoughtworks:

PR #692, opened 2019-07-15
Wrap console log && support --no-log

rsun-thoughtworks:
Fix issue #690 

Provide an option --no-log to disable all log from taiko itself, but still, allow user's script to use console.log

```sh
# will disable logs from within taiko itself
$ taiko user.js --no-log
```

Wrap console log && support --no-log

### DIFF
--- a/bin/runFile.js
+++ b/bin/runFile.js
@@ -1,6 +1,7 @@
 const path = require('path');
 const util = require('util');
 const recorder = require('../recorder');
+const { safeLog } = require('../lib/util');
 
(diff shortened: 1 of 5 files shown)


NivedhaSenthil:
silencing logs from REPL is not needed.. can be done only here to stop logging from runner here https://github.com/getgauge/taiko/blob/babdc48cf1328cb0a3ba0402d377e4752f90b465/bin/runFile.js#L46

NivedhaSenthil:
can you please look at the failures ? gotta check why few jobs are passing irrespective of the failure

Issue #688, opened 2019-07-15
additional write option to hide text from output


Brodan:
### Expected behavior
The `write` page action ([docs](https://taiko.gauge.org/#write)) should accept a boolean option (`hidden` for example) that explicit prevents the text being written to screen.

### Actual behavior
Currently there exists logic ([here](https://github.com/getgauge/taiko/blob/master/lib/taiko.js#L1043)) which will output `*****` if the active element's `isPassword` is True. However, there may be additional types of text other than passwords that one would not want written to terminal/logged.

Brodan:
closing this as #709 has been merged. thank you for the quick review and merge!

Debashis9012:
It has been verified and found fixed with the version:
```version: 1.0.1 (Chromium: 76.0.3803.0) 7b4db1995facc68d8a93cd62fa5ed6e29e37a171```

PR #709, opened 2019-07-22
add explicit hidden option to write command

Brodan:
#688.

In some cases (not just password inputs) you may not want the text from a `write` command to be written to the log output. 

This PR adds a `hidden` option to the `write` command which will output `*****` instead of the give text to output. 

Example:
```
brodan@~/taiko (master)$ node bin/taiko.js 
> openBrowser();
 ✔ Browser opened
> goto("https://www.google.com/");
 ✔ Navigated to URL https://www.google.com/
> write("This will not be logged.", {hidden:true});
 ✔ Wrote ***** into the focused element.
```


lib/taiko.js
```diff
@@ -1018,6 +1018,7 @@ module.exports.focus = async (selector, options = {}) => {
  * @param {number} [options.waitForStart=100] - wait for navigation to start. Accepts time in milliseconds.
  * @param {number} [options.timeout=30000] - DEPRECATED timeout option, use navigationTimeout instead.
  * @param {number} [options.navigationTimeout=30000] - Navigation timeout value in milliseconds for navigation after click.
+ * @param {boolean} [options.hidden=false] - Prevent given text from being written to log output.
```
nehashri: The options.hidden has a different meaning in some APIs. It means that the element to perform the action is hidden.
To avoid confusion, its better we use a different name here. In my opinion, `options.hideText` is a good name to use.

Brodan: @nehashri thanks for the quick look! I struggled to come up with a name for it and was definitely iffy on `hidden`. I'll change it to your suggestion!

Brodan:
Looks like the CI build timed out, any way to restart it?

nehashri:
@Brodan I am looking at why the builds are failing.

PR #724, opened 2019-07-30
Log and throw error if some error happened while dispatching event

luciferankon:


lib/taiko.js
```diff
@@ -1868,15 +1868,23 @@ function dropDown(attrValuePairs, ...args) {
                     this.selectedIndex = value.index;
                     found_value = true;
                     let event = new Event('change');
-                    this.dispatchEvent(event);
+                    try{
+                        this.dispatchEvent(event);
+                    }catch(e){
+                        return 'Error occured while dispatching event ' + e;
```
zabil: This error message is not user friendly. Plus it gives a dump of the stack.
Please handle this is a way that a user can understand and take action. 

Also do not throw errors as a string if you want to identify this throw a generic appError
https://www.npmjs.com/package/apperror

lib/taiko.js
```diff
@@ -1885,6 +1893,7 @@ function dropDown(attrValuePairs, ...args) {
             const options = setNavigationOptions({});
             await doActionAwaitingNavigation(options, async () => {
                 const result = await runtimeHandler.runtimeCallFunctionOn(selectBox, null, { nodeId: nodeId, arg: value });
+                if(typeof result.result.value != 'boolean') throw new Error(result.result.value);
```
zabil: This comparison is wrong again refer 
https://www.npmjs.com/package/apperror to see if this can be handled as an error and not using string comparison

lib/taiko.js
```diff
@@ -1885,6 +1893,7 @@ function dropDown(attrValuePairs, ...args) {
             const options = setNavigationOptions({});
             await doActionAwaitingNavigation(options, async () => {
                 const result = await runtimeHandler.runtimeCallFunctionOn(selectBox, null, { nodeId: nodeId, arg: value });
+                if(result.result.value instanceof Object) throw new Error(result.result.value);
```
zabil: Please be specific about the type you compare

For e.g.
```
if(result.result.value.stack instanceof [whatever the instance of e is]) ...
```
Make sure you do a proper null check to see if there's a message and stack

Also 
```
new Error(result.result.value) 
```

will throw the error as follows

```
   throw new Error({message: 'This is a message', 'stack': e});
   ^

Error: [object Object]
    at Object.<anonymous> (/Users/zabilcm/Projects/sample/sample.js:23:10)
```

As you can see it has gobbled up the error to fix this you will have to throw it as

```
throw result.result.value.stack
```

or 

```
throw new Error(result.result.message)
```

Is there some kind of test for this?

lib/taiko.js
```diff
@@ -1868,15 +1868,23 @@ function dropDown(attrValuePairs, ...args) {
...
                 for (var i = 0; i < this.options.length; i++) {
                     if (this.options[i].text === value || this.options[i].value === value) {
                         this.selectedIndex = i;
                         found_value = true;
                         let event = new Event('change');
-                        this.dispatchEvent(event);
+                        try{
+                            this.dispatchEvent(event);
+                        }catch(e){
+                            return {'message': 'Error occured while dispatching event ','stack': e};
```
zabil: Unwanted space at the end of the message
modify to
```
{'message': 'Error occured while dispatching event','stack': e}
```

luciferankon:
> @luciferankon We should include the actual error message not only the message created by us (Error occurred while dispatching event).
What should be the structure of reporting the error then?

lib/taiko.js
```diff
@@ -2426,75 +2426,63 @@ module.exports.comboBox = (attrValuePairs, ...args) => {
...
+                }
+                for (var i = 0; i < this.options.length; i++) {
+                    if (this.options[i].text === value || this.options[i].value === value) {
+                        this.selectedIndex = i;
+                        found_value = true;
+                        let event = new Event('change');
+                        try{
+                            this.dispatchEvent(event);
+                        }catch(e){
+                            return {'message': 'Error occured while dispatching event','stack': e};
```
negiDharmendra: The function being executed on the browser console can not return non primitive values unless [returnByValue](https://github.com/getgauge/taiko/blob/14d7793ce9d59ace9a9ad70fbf9750b56672ede5/lib/handlers/runtimeHandler.js#L44) is set to true.

luciferankon: @negiDharmendra Updated the logging structure. Please check it once.

Issue #733, opened 2019-08-05
write api should throw proper error message


Debashis9012:
### Expected behavior
`write` api should throw a proper error message as we have not mention in which textBox we have to perform write operation.

### Actual behavior
```
> write("Hello",textBox())
 ✔ Wrote Hello into the
```
### Steps to reproduce
1. openBrowser()
2. goto("file:///Users/xyz/Desktop/writeApiTest.html");
3. `write("Hello",textBox())`

Kindly find the below html snippet:
```
<html>
<body>
<div>
           <form name="inputTypeText">
               <div>
                   <input type="text" disabled='true' id='disabled-input' autofocus>initially disabled input-type-text</input>
               </div>
           </form>
           <script type="text/javascript">
               setTimeout( () => {
                   document.getElementById('disabled-input').disabled = false;
               }, 200);
           </script>
       </div>
</body>
</html>

```

### Versions
```
version: 1.0.1 (Chromium: 76.0.3803.0) 5ada00f72c218606b9681f18966987ac606bcaba
```

Debashis9012:
This issue has been verified and found fixed. 
Tested version: ```Version: 1.0.1 (Chromium: 76.0.3803.0) f30e18e5b28f9d57b39d06038f075dd0ccc7d4e9```

PR #736, opened 2019-08-07
#fix733

saikrishna321:
Address: #733


saikrishna321:
@NivedhaSenthil @BugDiver

PR #785, opened 2019-09-01
text() should string when single value is found

saikrishna321:
Current:

```
$('#foo').text() returns array even when single text is found
```

Fix:

```
$('#foo').text() returns 'string'  when single text is found
$('#foo').text() retuns 'array' when multiple text is found
```

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -3432,7 +3432,7 @@ const selectorText = get => {
         })).value,
       );
     }
-    return texts;
+    return texts.length > 1 ? texts : texts[0];
   };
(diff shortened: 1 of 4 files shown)


zabil:
It's better to not change this behaviour for the following reasons

* This is a part of 1.0 and will be a breaking change
* The API must have predictable return types. It should either only return string i.e the first match and use another api for multiple matches.

So I recommend leaving the API as it is.

nehashri:
We decided to go with https://github.com/getgauge/taiko/issues/808.

PR #792, opened 2019-09-04
Exposed api isDisabled

luciferankon:


nehashri:
We will have to revisit this after we fix https://github.com/getgauge/taiko/issues/808. Closing this for now.

PR #807, opened 2019-09-11
fix for open browser parameter type

andriikobtsev:


lib/taiko.js
```diff
@@ -191,6 +191,9 @@ eventHandler.addListener('targetCreated', async newTarget => {
  * @returns {Promise}
  */
 module.exports.openBrowser = async (options = { headless: true }) => {
+  if (typeof options !== 'object'){
```
luciferankon: It's better to check with string itself, As arrays, maps, etc are a type of object in javascript.

andriikobtsev: It is fair point regarding objects in javascript. However, in this particular case I thought it should be ok to check for object. Investigation shown the problem actually appears with strings, numbers, boolean. And works with objects (objects, arrays, maps). So the solution with checking for string can not help to eliminate the problem entirely. I spent some time to dig really deep to find the root cause, but without the result so far(ended up in browserFetcher.waitForWSEndpoint). I will be happy to discuss further or can your opinion regarding my comment. Thank you.

PR #829, opened 2019-09-23
Add a return at dialog method. Then developers will be able to remove…

felipeespitalher:
Add a return to the dialog's method. Then developers will be able to remove nonused listeners and avoid memory leak


felipeespitalher:

PR #833, opened 2019-09-25
Fix to get button with child element

shubhamsc:


test/unit-tests/textMatch.test.js
```diff
@@ -207,48 +210,6 @@ describe('match', () => {
...
-    describe('match text for type and paragraph', () => {
-      it('test exact match for type', async () => {
-        expect(await text('text').exists()).to.be.true;
-        expect(await text('text').get()).to.have.lengthOf(8);
-      });
-      it('test contains match for type and text', async () => {
-        expect(await text('tex').exists()).to.be.true;
-        expect(await text('tex').get()).to.have.lengthOf(11);
-      });
-    });
```
negiDharmendra: Why we had to remove these tests.

lib/elementSearch.js
```diff
@@ -97,6 +97,13 @@ async function matchContainsElements(
       translateTo,
       xpathText,
     );
+  if (e === 'button' && (!elements || !elements.length))
```
negiDharmendra: In my opinion we should rename the parameter `e` to `tagName`.

test/unit-tests/textMatch.test.js
shubhamsc: These are duplicate tests, written in two places. See line 145-186 and line 210-251
https://github.com/getgauge/taiko/blob/master/test/unit-tests/textMatch.test.js#L145
https://github.com/getgauge/taiko/blob/master/test/unit-tests/textMatch.test.js#L210

lib/elementSearch.js
shubhamsc: I agree we are using variable `e` in many places in this function and in other function calls.
This should be `tagName`, which need to replace everywhere to avoid confusion.

negiDharmendra: I think we should rename it as part of this PR.

PR #850, opened 2019-09-30
Provide more details in dot api

### LINKED ISSUE #834: .api with the name of the api should have all detailed information
### Expected behavior
.api `"api name"` should have all detailed information.

Ex: `.api inputField`
As we have deprecated `inputField` api to `textBox`, it should show the deprecation message.

It should have `parameters`, `deprecation note`, `return type for an api`.
Its better to have these information which will be useful for the user.

### Actual behavior

```
> .api inputField

This selector lets you identify an input field on a web page with label or attribute and value pairs and proximity selectors.

Example:
	await focus(inputField({'id':'name'})
	await inputField({'id': 'name'}).exists()
	await inputField({id:'inputFieldId'},below('text')).exists()
	await inputField(below('text')).exists()
```
### Versions
```
Version: 1.0.2 (Chromium: 76.0.3803.0) c9fdf2959cce1b94e6ed6964aecaca505c5cf43d
```

M4rk9696:
Fix #834

### DIFF
--- a/lib/repl/repl.js
+++ b/lib/repl/repl.js
@@ -4,6 +4,7 @@ const util = require('util');
 const { aEval } = require('./awaitEval');
 const { initSearch } = require('./repl-search');
 const { removeQuotes, symbols } = require('../util');
+const { EOL } = require('os');
 const funcs = {};
 let commands = [];
 const stringColor = util.inspect.styles.string;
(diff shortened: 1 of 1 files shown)


lib/repl/repl.js
```diff
@@ -266,7 +266,24 @@ function displayUsageFor(name) {
     return;
   }
   console.log();
+  if (e.deprecated) {
+    console.log(`DEPRECATED ${desc(e.deprecated)}`);
+    console.log();
```
BugDiver: Any specific reason for empty `console.log`?
If it is just to achieve a new line, consider using '\n'. If worried about os specific new lines consider using `require('os').EOL`.

lib/repl/repl.js
```diff
@@ -266,7 +266,24 @@ function displayUsageFor(name) {
...
+    console.log(e.params.length > 1 ? 'Parameters:' : 'Parameter:');
+    console.log();
+    params(e.params);
+  }
+  if (e.returns) {
+    console.log();
+    e.returns.map(e => {
+      console.log(`Returns: ${type(e.type)} ${(e.description.type ? desc(e.description) : e.description)}`);
+    })
+    console.log();
```
BugDiver: Empty `console.log`?

lib/repl/repl.js
M4rk9696: I was reusing what was previously in place. `EOL` is a better fit as suggested.

M4rk9696:
@BugDiver There seems to be a lot of files not linted properly, running `lint` changes a lot of files. Should I raise a separate PR with all lints?

BugDiver:
I guess that would be better.

saikrishna321:
@BugDiver @M4rk9696 There is a pre-commit hook that does linting.

https://github.com/getgauge/taiko/blob/master/package.json#L30-L33

M4rk9696:
@saikrishna321 I think it is better to have a lint stage in CI as well, as a hook can be skipped.

Issue #766, opened 2019-08-26
text API should return only exact match


saikrishna321:
text API returns true for partial match text.

`<div>Editable Demo</div>
`

TextAPI

```
text('demo').exists() // true
text('Editable').exists() // true
text('editable demo').exists() // true
```

Version: Taiko master

Apoorva-GA:
@saikrishna321 Text api does exact match first and if no matches are found it checks for contains and this behaviour was introduced intentionally. 

Note : Previously we had a separate api `contains` for the partial text match which we have removed now and `text` api does the job now.

saikrishna321:
> Text api does exact match first and if no matches are found it checks for contains 

This does not sound correct IMO. I would want my tests to fail if I don't have an exact match

Apoorva-GA:
Discussion update :
`text` API to take  a strategy option. Something like exact match, starts with etc.

Extension to this would be text api to take in a function. Logged as a separate [issue](https://github.com/getgauge/taiko/issues/773).

saikrishna321:
Thank's @Apoorva-GA

zabil:
Passing a strategy of options breaks Taiko's flat and simple API. 
Plus it won't cater to all the selection options.

The simplest way is to accept a regular expression as a parameter. For example to select the exact string that would be

```
text(/something/).exists()
```

This will open up the API to much more options without adding a parameter or reading up the documentation.

nehashri:
@zabil Regex match of text is not supported with XPath 1.0. Chrome uses XPath 1.0, so it will be hard to support regex match of text as of now.

zabil:
Since there's no support for Regex. We can return the value of the matched element instead. This can be used for an assertion. For e.g.

```
<div>Editable Demo</div>
```

```
text('demo').exists() // true
text('demo').value() // Editable Demo
```

saikrishna321:
> text('demo').value() // Editable Demo

@zabil IMHO This approach is not a very friendly. The user would always need to pick the value and then assert. 

I have created #1122 were `text` API will take `exactMatch` as option.

zabil:
@saikrishna321 Did you add documentation for the API as a part of the this pull request. I don't seem to find it when I generate the docs.

If not can you add documentation and close this  issue?

saikrishna321:
Sorry, my bad! Will do that

PR #855, opened 2019-10-01
Elements Search with ContainsText & StartsWithText

jayandran-Sampath:
Created two method to enable element search using Element Contains some text & Elements Starts with some text

Example : 
containsText('some text')
startsWithText('some text')

Since this is my Initial PR, feel free to provide feedbacks on improvement factors as well.


saikrishna321:
@jayandran-Sampath `text` API looks for an exact match if not found it moves to contains. Can we enrich this PR to have api which does exactMatch - https://github.com/getgauge/taiko/issues/766

PR #859, opened 2019-10-03
enable reload api to accept ignoreCache and deprecate url arg

saikrishna321:


lib/taiko.js
```diff
@@ -2449,66 +2454,96 @@ module.exports.comboBox = (attrValuePairs, ...args) => {
...
-                        try{
-                            this.dispatchEvent(event);
-                        }catch(e){
-                            return {'error': 'Error occured while dispatching event','stack': e.stack};
-                        }
-                        break;
-                    }
-                }
-                return found_value;
+  validate();
```
nehashri: I think this is just a lint change. Can we not commit it with this PR?

saikrishna321: Agree!

nehashri:
Merging this PR for now, Will add tests to this later.

Issue #811, opened 2019-09-13
Add option `selectHiddenElement` to selectors


nehashri:
This option exists for the fileField selector. Similar option should be exposed to other selectors as well.

List of selectors should be changed:
- [ ] $
- [x] image
- [x] link
- [ ] listItem
- [x] button
- [ ] textBox
- [ ] dropDown
- [x] checkBox
- [x] radioButton
- [ ] text

nehashri:
Related https://github.com/getgauge/taiko/issues/630

jayandran-Sampath:
@sathiyaraj45 - Working on this issue

sivaranjanidev:
Me and @nandhae are working for `image` element

jayandran-Sampath:
Im Working on Button element

nehashri:
Not all selectors have this option enabled.

saikrishna321:
@nehashri Can we please mark the issue for the ones completed. I did see for `listView` and `textbox` that is missing

harrymt:
Could we have this functionality for the [`fileField`](https://docs.taiko.dev/#filefield) selector please

PR #866, opened 2019-10-05
Add option `selectHiddenElement` - To Link Selector

jayandran-Sampath:
Add option `selectHiddenElement` to Link Selector, myself & @sathiyaraj worked on enabled Hidden Element Search for Link Selector, If we are good with this, i will do the same for remaining selectors

Part of #811 - link


lib/taiko.js
```diff
@@ -2042,6 +2043,70 @@ module.exports.link = (attrValuePairs, ...args) => {
...
+  assertType(text);
+  let nbspChar = String.fromCharCode(160);
+  let textToTranslate = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ' + nbspChar;
+  let translateTo = 'abcdefghijklmnopqrstuvwxyz' + ' ';
+  let xpathText = `translate(normalize-space(${xpath(
+      text,
+  )}), "${textToTranslate}", "${translateTo}")`;
+  const get = async () => await handleRelativeSearch(await $$xpath(
+      `//text()[contains(translate(normalize-space(string()), "${textToTranslate}", "${translateTo}"), ${xpathText})]`,
+  ),args);
```
saikrishna321: Can't we use the exists fn - https://github.com/getgauge/taiko/pull/866/files#diff-fa054c0cd359737bbdc57062a6fee95aR122

saikrishna321: Can the logic be moved to `elementSearch.js`? @nehashri Thoughts?

lib/taiko.js
```diff
@@ -2042,6 +2043,70 @@ module.exports.link = (attrValuePairs, ...args) => {
...
+    assertType(text);
+    let nbspChar = String.fromCharCode(160);
+    let textToTranslate = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ' + nbspChar;
+    let translateTo = 'abcdefghijklmnopqrstuvwxyz' + ' ';
+    let xpathText = `translate(normalize-space(${xpath(
+        text,
+    )}), "${textToTranslate}", "${translateTo}")`;
+    const get = async () => await handleRelativeSearch(await $$xpath(
+        `//text()[starts-with(translate(normalize-space(string()), "${textToTranslate}", "${translateTo}"), ${xpathText})]`,
+    ),args);
```
saikrishna321: Same as above!

lib/taiko.js
```diff
@@ -2042,6 +2043,70 @@ module.exports.link = (attrValuePairs, ...args) => {
...
+    return {
+        get: getIfExists(get),
+        exists: exists(getIfExists(get), descEvent),
+        description: `Element starts with text "${text}"`,
+        text: selectorText(get),
+    };
+};
```
negiDharmendra: @jayandran-Sampath  Please separate these changes from this PR.

jayandran-Sampath:
@negiDharmendra @saikrishna321 Sorry my bad, i raised this PR on top of existing PR, I will work on it properly and then raise it back. Thanks

lib/elementSearch.js
```diff
@@ -95,46 +95,54 @@ async function matchContainsElements(
...
-  xpathText,
+  xpathText,options,
 ) {
-  return await $$xpath(
-    `//text()[translate(normalize-space(string()), "${textToTranslate}", "${translateTo}")=${xpathText}]`,
+  return options ? await $$xpath(
+    `//text()[translate(normalize-space(string()), "${textToTranslate}", "${translateTo}")=${xpathText}]`,options.selectHiddenElement
+  ) : await $$xpath(
+      `//text()[translate(normalize-space(string()), "${textToTranslate}", "${translateTo}")=${xpathText}]`,
   );
```
negiDharmendra: We don't need this check here as [$$xpath](https://github.com/getgauge/taiko/blob/73243dee5cbddb9e69ae30b4391b8d26a288f20e/lib/elementSearch.js#L249) already takes care of this scenario. I.M.O we should just pass the `selectHiddenElement` option to the `$$xpath` function.

jayandran-Sampath: @negiDharmendra thanks, i have tried without this condition but getting exception, thats why  handled it

Issue #612, opened 2019-06-12
Support for html 5 input types


Apoorva-GA:
Support the following input types.
- [x] color
- [ ] date
- [ ] datetime-local
- [x] image
- [ ] month
- [x] range
- [ ] time
- [ ] week

wkoch:
I'm working on this issue.

wkoch:
Judging by [MDN Dev Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/image), I think "input type image" should be detected as a button instead of a textBox, any thoughts?

zabil:
>  "input type image" should be detected as a button instead of a textBox

Yes that's right.

wkoch:
I implemented input image and extended the existing tests and all tests pass.

I'm having trouble with the other input types, just extending the tests fail for all new input types (always return empty). I think there's some conflict between the current tests and the html5 inputs since I'm using the same changes in a modded Taiko.js file in a personal project and I have tests working on input time using textBox.

I'll try writing a set of new tests specifically for the new input types this weekend just to see if I can figure out the problem.

nehashri:
@wkoch I suggest you raise PRs for the ones you have finished, you can always raise more than one PR for this issue. This will help keep the code upto date and also avoid long PRs.

Apoorva-GA:
Support for `image` available in commit f3b7aafcede6f0185c06ac23f85451f497b4951a.

Debashis9012:
verified support for html5 input type image. Taiko able to perform click operation 
for input type image webelement.
Tested version: `afd215c9be7633fde970f2e464b3d144517bf638`

o-x-y-g-e-n:
All the ones left for implementation falls under TextBox... right?

wkoch:
> 
> 
> All the ones left for implementation falls under TextBox... right?

Yes.

o-x-y-g-e-n:
@wkoch If I may know, are you still working on this issue? If not, I might consider this great opportunity to try to implement the rest of them. Thanks.

wkoch:
> @wkoch If I may know, are you still working on this issue? If not, I might consider this great opportunity to try to implement the rest of them. Thanks.

Please do. I'm also needing this functionality.

o-x-y-g-e-n:
@zabil , considering 'Range' input type, we do want to expose user with 'min', 'max' and setting the value of the range input.. right?

zabil:
[...]

PR #901, opened 2019-10-26
Added input image as a button

wkoch:
I Added the html5 input image as a type of button since it [acts like one](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/image)).

I just extended the current button tests and everything works.

Partial fix for Issue #612


Apoorva-GA:
@wkoch Thanks for thr PR!

PR #899, opened 2019-10-24
Add support for 'selectHiddenElement' to checkbox selector

nandhae:
**What this PR does**
Adds support to 'selectHiddenElement' for 'checkbox' selector as part of #811 

Adds unit tests for the functionality.

PR #916, opened 2019-10-31
Add support to find hidden element for RadioButton

Srilalitha1310:
**What this PR does**

1. Part of #811 
2. Follows same approach as #899


Srilalitha1310:
@nehashri When i gave the style to hide the radio button inside the tag, unit tests failed in Windows, but passed in MacOS and Linux. I modified it by moving the logic inside script and it worked and passed.

Issue #821, opened 2019-09-17
add wildcards to switchTo command


sarut0bi:
Hi it seem that curently it is impossible to use wildcards in switchTo function.
my use case is the following:
* The title of the page is language dependant ...
* url contains uuid :(

Is there a way to add regex rule like it have been already done for intercept api?

zabil:
While implementing this modify the api to take a regex argument for example

```
switchTo(/http(s?):\/\/(www?).google.(com|co.in|co.uk)/)
```

o-x-y-g-e-n:
@zabil, @Apoorva-GA  would this syntax work? If starts with r'(.....) consider the inner text as regex.
`switchTo('r(http(s?):\/\/(www?).google.(com|co.in|co.uk))')`

o-x-y-g-e-n:
I would like to take this opportunity to implement this.

sriv:
@o-x-y-g-e-n - please go ahead and work on this. You may use this issue for any discussions.

zabil:
`switchTo` now supports regex

PR #968, opened 2019-12-12
#821 Added Regex support for SwitchTo & CloseTab methods.

o-x-y-g-e-n:
- Added support for regex in switchTo & CloseTab.
- Added unit tests for the changes made.


lib/taiko.js
```diff
@@ -341,23 +355,52 @@ function getEventProxy(target) {
...
  *
  * @example
- * await switchTo('https://taiko.gauge.org/') # switch using URL
- * await switchTo('Taiko') # switch using Title
+ * # switch using URL
+ * await switchTo('https://taiko.gauge.org/')
+ * # switch using Title
+ * await switchTo('Taiko')
+ * # switch using regex URL
+ * await swithTo(regex(/http(s?):\/\/(www?).google.(com|co.in|co.uk)/))
```
zabil: There's a typo this should be `switchTo`.

The idea is to make this api as transparent to the user. Instead of introducing a `regex` method can't we simply check the parameter type and identify this as a regex?

For example (simplified version) check if a parameter is a regex

```
function switchTo(param) {
 if(param instanceof RegExp) {
 } else {}
}
```

If this check is internal then the user does not have to export the regex method. The can just specify the parameter and this api can be.

```
swithTo(/http(s?):\/\/(www?).google.(com|co.in|co.uk)/)
```

o-x-y-g-e-n: Ofcourse. I thought of that too. But now suppose that you want to add this regex functionality into other functions. I think regex() will provide a base for all of the functionalities that accepts regex. 

Consider it as an `before()` kinda thing. You can pre-process your regex before doing the real stuff.

Let me know if this sounds good. Otherwise I don't mind removing it and simply parsing the regex.

Thanks.

zabil: I agree about the re-suability. But we could do the pre-processing inside the `switchTo` and `closeTab`

The `regex` method here is very different from the other api's as it does not perform an action on the browser. Taiko api's are trying to make the api very specific to actions on the browser.

Also can't what this method does 

```
module.exports.regex = async regexString => {
   var isValid = true;
   var myRegexObject = null;
   try {
     myRegexObject = new RegExp(regexString);
   } catch (e) {
     isValid = false;
   }
   if (!isValid) {
     throw new TypeError(`The "regexString" must be a valid regex`);
   } else {
     return myRegexObject;
   }
 };
```

simply be 

```
regexString instanceof RegExp
```

When the user passes `/test/`? Else, just treat it as a string and use that for matching. This way a tester does not have to remember to use the extra api. When it is an instance of `RegExp` it should be valid to use and match.

Again  the api could be

```
await closeTab(/http(s?):\/\/(www?).google.(com|co.in|co.uk)/)
switchTo(/http(s?):\/\/(www?).google.(com|co.in|co.uk)/)
```

with no imports or extra api for the user to learn. We are always trying to make it easier for the user. Hope this makes sense. Sorry for pushing this!

o-x-y-g-e-n: Gotcha! I will push with the changes. :)

o-x-y-g-e-n:
@zabil Now switchTo() & closeTab() takes regexString directly!

lib/taiko.js
```diff
@@ -345,33 +359,53 @@ module.exports.client = () => clientProxy;
...
+  }
   const targets = await targetHandler.getCriTargets(
     targetUrl,
     currentHost,
     currentPort,
   );
   if (targets.matching.length === 0) {
-    throw new Error('No target with given URL/Title found.');
+    throw new Error(
+      `No target with given ${isRegexText} URL/Title found.`,
```
zabil: This can be a generic message to avoid the conditions on `isRegexText` 
`No tab(s) matching ${tartgetURL} found`

lib/taiko.js
```diff
@@ -345,33 +359,53 @@ module.exports.client = () => clientProxy;
...
+    );
   }
   await connect_to_cri(targets.matching[0]);
   await dom.getDocument();
-  descEvent.emit('success', 'Switched to tab with URL ' + targetUrl);
+  descEvent.emit(
+    'success',
+    `Switched to tab with${isRegexText} URL ` + targetUrl,
```
zabil: This can be
`Switched to tab matching ${targetUrl}`

lib/taiko.js
```diff
@@ -582,15 +632,18 @@ module.exports.closeTab = async targetUrl => {
...
     await connect_to_cri(others[0]);
     await dom.getDocument();
   }
   let message = targetUrl
-    ? `Closed all tabs with URL ${targetUrl}`
-    : `Closed current tab with URL ${closedTabUrl}`;
+    ? `Closed all tabs with${isRegexText} URL ${targetUrl}`
+    : `Closed current tab with${isRegexText} URL ${closedTabUrl}`;
   descEvent.emit('success', message);
 };
```
zabil: * The argument check for `targetUrl` can be removed as it happens in `getCriTargets`
* Lot of if conditions to handle the display message. It's better to make it generic 

Here are some modifications I did to the whole call chain the changes will be similar for `switchTo`

```
module.exports.closeTab = async targetUrl => {
  validate();

  const { matching, others } = await targetHandler.getCriTargets(
    targetUrl,
    currentHost,
    currentPort,
  );
  if (!others.length) {
    await _closeBrowser();
    descEvent.emit('success', 'Closing last target and browser.');
    return;
  }
  if (!matching.length) {
    throw new Error('No target with given URL/Title/Regex found.');
  }
  let currentUrl = await currentURL();
  let closedTabUrl;
  for (let target of matching) {
    closedTabUrl = target.url;
    await cri.Close({
      host: currentHost,
      port: currentPort,
      id: target.id,
    });
  }
  if (
    !targetHandler.isMatchingUrl(others[0], currentUrl) &&
    !targetHandler.isMatchingRegex(others[0], currentUrl)
  ) {
    _client.removeAllListeners();
    _client = null;
    await connect_to_cri(others[0]);
    await dom.getDocument();
  }
  let message = `Closed tab(s) matching ${targetUrl || closedTabUrl }`
  descEvent.emit('success', message);
};
```

```
const getCriTargets = async function(targetObject, host, port) {
  let targets = await cri.List({ host: host, port: port });
  let pages = targets.filter(target => target.type === 'page');
  let response = {
    matching: pages.slice(0, 1),
    others: pages.slice(1),
  };
  if (targetObject) {
    response = { matching: [], others: [] };
    for (const target of targets) {
      if (
        (targetObject instanceof RegExp && isMatchingRegex(target, targetObject)) ||
        isMatchingUrl(target, targetObject)
      ) {
        response.matching.push(target);
      } else {
        response.others.push(target);
      }
    }
  }
  return response;
};
```

```
const isMatchingRegex = function(target, targetRegex) {
  const parsedUrl = url.parse(target.url, true);
  const host = parsedUrl.host ? parsedUrl.host : '';
  const urlPath = path.join(
    host,
    trimCharLeft(parsedUrl.pathname, '/'),
  );
  const urlHrefPath = parsedUrl.protocol
    .concat('//')
    .concat(trimCharLeft(host, '/'));

  return (
    target.title.match(targetRegex) ||
    urlPath.match(targetRegex) ||
    parsedUrl.href.match(targetRegex) ||
    urlHrefPath.match(targetRegex)
  );
};

const isMatchingUrl = function(target, targetUrl) {
  const parsedUrl = url.parse(target.url, true);
  const parsedTargetUrl = url.parse(targetUrl, true);

  const targetHost = parsedTargetUrl.host ? parsedTargetUrl.host : '';
  const host = parsedUrl.host ? parsedUrl.host : '';

  const targetUrlPath = path.join(
    targetHost,
    trimCharLeft(parsedTargetUrl.pathname, '/'),
  );
  const urlPath = path.join(
    host,
    trimCharLeft(parsedUrl.pathname, '/'),
  );

  target.url = handleUrlRedirection(target.url);
  return (
    target.title === escapeHtml(targetUrl) ||
    urlPath === (targetUrlPath || targetUrl) ||
    parsedUrl.href === targetUrl
  );
};
```

Notice that there are no parameter condition checks in `isMatchingUrl` and `isMatchingRegex` The responsibility can be at `getCRITargets` i.e. these lines

```
     if (
        (targetObject instanceof RegExp && isMatchingRegex(target, targetObject)) ||
        isMatchingUrl(target, targetObject)
      )
```

One of the modified test will be as follows. 

```
it('should close the current tab and switch to last active target if no url given', async () => {
    _targets.matching.push({
      id: '1',
      type: 'page',
      url: 'https://flipkart.com',
    });
    _targets.others.push({
      id: '2',
      type: 'page',
      url: 'https://flipkart.com',
    });
    _targets.others.push({
      id: '3',
      type: 'page',
      url: 'https://amazon.com',
    });
    currentURL = 'https://amazon.com';
    _isMatchUrl = false;
    _isMatchRegex = false;
    let validatePromise = validateEmitterEvent(
      'success',
      'Closed tab(s) matching https://flipkart.com',
    );
    await taiko.closeTab();
    await validatePromise;
    expect(currentTarget.url).to.be.eql('https://flipkart.com');
  });
```

lib/taiko.js
```diff
@@ -345,33 +359,53 @@ module.exports.client = () => clientProxy;
...
-      'The "targetUrl" argument must be of type string. Received type ' +
+      'The "targetUrl" argument must be of type string or regex. Received type ' +
         t,
     );
   }
+  if (q) {
+    targetUrl = new RegExp(targetUrl);
+    isRegexText = ' Regex';
+  }
   const targets = await targetHandler.getCriTargets(
```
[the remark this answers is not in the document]
o-x-y-g-e-n: @zabil , Actually `/regexString instanceof RegExp/` always returns false in our case i.e `targetObject instanceof RegExp` is always false.

This is the reason i guess : https://stackoverflow.com/questions/4339288/typeof-for-regexp#comment4724685_4339350

That's the reason i had to do something like this :
https://github.com/o-x-y-g-e-n/taiko/blob/738fb2c2d47cf3ae87273f5e1abc86a9c16a9098/lib/taiko.js#L371-L374

I am not able to know the exact reason as per what is happening, so the workaround would be instead of `/regexString/ instanceof RegExp` , the condition would be `Object.prototype.toString.call(targetUrl).includes('RegExp')`

lib/taiko.js
o-x-y-g-e-n: I see you have added this condition 
`if (
        (targetObject instanceof RegExp) && isMatchingRegex(target, targetObject)) ||
        isMatchingUrl(target, targetObject)
      )`

But this doesn't work always.
Suppose that I pass something like,
`
openBrowser()
goto('google.com')
openTab('github.com')
switchTo(/Go*gle/)
`
The stacktrace would be something like.
1. Go to `isMatchingRegex("github.com",/Go*gle/)`.
  The above statement will return false. and hence it will go and try executing isMatchingUrl().
2. Go to `isMatchingUrl('github.com',/Go*gle/)`
  For this method, the below line will cause exception.
`const parsedTargetUrl = url.parse(targetUrl, true);`

We are also looping through all the tabs open in the code. Hence, the check in both isMatchingUrl() & isMatchingRegex() is required. Please let me know what you think.

o-x-y-g-e-n:
@zabil Thank you for taking the time and effort for the review. Added a few points and clarifications from my side. Please let me know what you think. Thanks.

lib/taiko.js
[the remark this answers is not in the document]
zabil: That is weird, good catch. I understand the need for this check.

I've added a few comments for making the error/info messages common to avoid conditions on generating them. I guess once that is done this PR should be good.

lib/taiko.js
```diff
@@ -553,13 +573,24 @@ module.exports.openTab = async (
...
+ * await closeTab(/http(s?):\/\/(www?).google.(com|co.in|co.uk)/)
  *
  * @param {string} [targetUrl=undefined] - URL/Page title of the tab to close.
  *
  * @returns {Promise}
  */
 module.exports.closeTab = async targetUrl => {
   validate();
+  if (
+    targetUrl != null &&
```
o-x-y-g-e-n: @zabil  Why is the null check important here? Because if we don't check it, then it will parse it into regex (/*/) and this will match to all the window , hence closing all the tabs & the window.

lib/taiko.js
```diff
@@ -346,33 +352,45 @@ module.exports.client = () => clientProxy;
...
  * @param {string} targetUrl - URL/Page title of the tab to switch.
  *
  * @returns {Promise}
  */
+
 module.exports.switchTo = async targetUrl => {
   validate();
-  let t = typeof targetUrl;
-  if (t !== 'string') {
+  if (
```
o-x-y-g-e-n: Keeping this check, because i saw a test case which verifies if switchTo() throws a TypeError if no argument is passed. Hence not touching it. If i had directly converted it into RegExp, then switchTo() would evaluate to switchTo(/*/) and this will just switch to the current tab.

o-x-y-g-e-n:
Modified the log messages. Also I have tried to reset the formatting changes as much as possible. Added a few comments. Please let me know what you think. Thanks.

test/unit-tests/handlers/targetHandler.test.js
```diff
@@ -78,5 +111,23 @@ describe('TargetHandler', () => {
       expect(targets.matching.length).to.be.equal(0);
       expect(targets.others.length).to.be.equal(3);
     });
+    it('should give the matching tab for regex Title', async () => {
```
o-x-y-g-e-n: Added this test case which will check this logic fails. Thanks for pointing it out. 👍 
`if (
        (targetObject instanceof RegExp && isMatchingRegex(target, targetObject)) ||
        isMatchingUrl(target, targetObject)
      )`

o-x-y-g-e-n:
> Great work. Will merge this post 1.0.3 release which will happen this week

I really want to get started implementing #612 . Please let me know what you think of this. Thank you.
https://github.com/getgauge/taiko/issues/612#issuecomment-565919212

nehashri:
@o-x-y-g-e-n There seems to be some conflicts still, can you please fix those?

o-x-y-g-e-n:
> @o-x-y-g-e-n There seems to be some conflicts still, can you please fix those?

Hello @nehashri . I tried but i am getting a lot of lint errors. Did anyone else face such issue? 

> @o-x-y-g-e-n There seems to be some conflicts still, can you please fix those?

There you go :)

Issue #809, opened 2019-09-13
Expose isVisible api for selectors


nehashri:
List of selectors to support isVisible:
- [ ] $
- [ ] image
- [ ] link
- [ ] listItem
- [ ] button
- [ ] fileField
- [ ] textBox
- [ ] dropDown
- [ ] checkBox
- [ ] radioButton
- [ ] text

o-x-y-g-e-n:
I would like to take the opportunity to work on this.

saikrishna321:
API 👇 Done

- [x] image
- [x] link

PR #973, opened 2019-12-18
add element visible check

saikrishna321:
Address #809


lib/elementWrapperUtils.js
```diff
@@ -122,6 +126,15 @@ const generateElementWrapper = function(
       if (!elems[0]) throw new Error(`${this.description} not found`);
       return await elems[0].text();
     },
+    isVisible: async function(retryInterval, retryTimeout) {
+      const elems = await this.elements();
+      if (elems.length < 1) {
+        descEvent.emit('success', 'Element not Visible');
```
nehashri: I think it's better to throw an error if no elements are found.

saikrishna321: I want to keep this consistent with `exists`. What u think?

PR #978, opened 2019-12-23
Added support for range input type (textBox API)

o-x-y-g-e-n:
- added support for range input type for textBox API.
- added test cases for the same.


test/unit-tests/textBox.test.js
```diff
@@ -342,6 +345,223 @@ describe(test_name, () => {
...
+          .true;
+      });
+
+      it('test value()', async () => {
+        expect(await textBox('With Wrapped Label').value()).to.equal(
+          inputType.testValue,
+        );
+        console.log(await textBox('With Wrapped Label').value());
+        await focus(textBox('With Wrapped Label'));
+        await press('ArrowRight');
```
o-x-y-g-e-n: 1st way to change the value of range.

test/unit-tests/textBox.test.js
```diff
@@ -342,6 +345,223 @@ describe(test_name, () => {
...
+        console.log(await textBox('With Wrapped Label').value());
+        expect(await textBox('With Wrapped Label').value()).to.equal(
+          (parseInt(inputType.testValue) + 1).toString(),
+        );
+        await evaluate(textBox('With Wrapped Label'), element => {
```
o-x-y-g-e-n: 2nd way to change the value of range.

test/unit-tests/textBox.test.js
```diff
@@ -342,6 +345,223 @@ describe(test_name, () => {
...
+
+    describe('attribute and value pair', () => {
+      it('test exists()', async () => {
+        expect(
+          await textBox({
+            id: inputType.name + 'WithLabelFor',
+          }).exists(),
+        ).to.be.true;
+        expect(
+          await textBox({
```
o-x-y-g-e-n: this is supported! 
`textBox({min:"10",max:"100"})`

PR #981, opened 2019-12-24
Added support for Html5 input elements

o-x-y-g-e-n:
Input Types covered (#612)
- date
- range
- color
- datetime-local
- month
- time
- week

Added test cases for the same. (sorry for totally irrelevant commit message  :))


o-x-y-g-e-n:
We will be adding two different API's for `range` & `color` but we are still supporting them in the `textbox` API.

Debashis9012:
#### Observation #1
Unable to handle input type range with the below code:
` write("0", textBox({name: `points`}))`
#### Excepted result: range point should be set to as "0" as per the statement.

#### Actual Result: 
` ✔ Wrote 0 into the text field[@name = concat('points', "")]` but the point still stands in default value there is no change in range.

#### HTML snippet
```
<!DOCTYPE html>
<html>
<body>

<h2>Range Field</h2>
<p>Depending on browser support:<br>The input type "range" can be displayed as a slider control.<p>

<form action="/action_page.php" method="get">
  Points:
  <input type="range" name="points" min="0" max="10">
  <input type="submit">
</form>

<p>
<b>Note:</b>
type="range" is not supported in Internet Explorer 9 and earlier versions.
</p>

</body>
</html> -->
```
#### Observation #2 
`textBox` api should be updated in the documentation as per the current changes.

Issue #1023, opened 2020-01-13
Is there a way to get the defaultConfig?


moogleberry:
### Expected behavior
Looking for a way to get the `defaultConfig` object that `taiko.js` and `config.js` use, and make it its values available to plugins and other custom addons.  

For example, if I build my own functionality that I want to use often in multiple tests, I want it to be able to know what the global `retryTimeout` and `retryInterval` values are.  

### Actual behavior
The only way I've found is to modify `taiko.js` to provide a getter function for it.  

### Versions
```
Taiko 1.0.3
```

negiDharmendra:
@moogleberry Right now there is no way to access the default config inside the `taiko-plugins`.
However if you want to set any of the default config you can do that by `setConfig` API.
**Example:**
```
setConfig( { observeTime: 3000});
      or
setConfig( { waitForNavigation: false, navigationTimeout: 3000});
```

moogleberry:
@negiDharmendra Would adding a `getConfig` in MR be welcome, or is that discouraged?

sriv:
IMO introducing `getConfig` makes sense.

Debashis9012:
Verified [PR 1036](https://github.com/getgauge/taiko/pull/1036)

PR #1036, opened 2020-01-15
Get config

moogleberry:
https://github.com/getgauge/taiko/issues/1023
Adds a `getConfig` function to `config.js` and exposes it in `taiko.js`


lib/taiko.js
```diff
@@ -3121,6 +3122,16 @@ getPlugins().forEach(pluginName => {
   }
 });
 
+/**
+ * Lets you read the global configurations.
+ *
+ * @example
+ * getConfig("retryInterval");
```
saikrishna321: Can you please add the other config param - 'navigationTimeout, observeTime, retryInterval, retryTimeout, observe, waitForNavigation, ignoreSSLErrors, headful, highlightOnAction';

lib/taiko.js
```diff
@@ -3121,6 +3122,16 @@ getPlugins().forEach(pluginName => {
...
+ *
+ * @param {String} paramName - Specifies the name of the paramter you want to get (optional).  If not specified, returns a shallow copy of the full global configuration. 
+*/
+module.exports.getConfig = getConfig;
```
saikrishna321: Unit test missing.

lib/taiko.js
moogleberry: Is this what you had in mind?

lib/taiko.js
moogleberry: @saikrishna321 Working on it, but is there a reason that in the `setConfig` test, the `allowedConfig` string is manually created instead of made from the keys of `defaultConfig`?  And should I do the same with the `getConfig` tests?

saikrishna321: @moogleberry 
> allowedConfig string is manually created instead of made from the keys of defaultConfig

I don't see a reason why not use `defaultConfig`.  Definitely a place to improve the test. It would be great if you can update the test pick from `defaultConfig`.

moogleberry: @saikrishna321 Ok did that

moogleberry:
@Debashis9012 any changed needed to this?

Debashis9012:
@moogleberry  Kindly allow me some time to look into it. I will get back to you soon.

Debashis9012:
Tested and verified, able to fetch the default Configuration by using `getConfig` api.

PR #1085, opened 2020-02-10
Interceptor Exact Match check before Reg Exp

### LINKED ISSUE #1083: Taiko Interceptors incorrectly handle Query Params
We have noticed that when trying to intercept a URL with a Query Param e.g.
`https://host.net/api/path?type=X` that the interceptor does not match as expected.

After doing some digging it seems that the interceptor matching logic does not escape special regexp characters in URLs properly e.g. `.` & `?`.
See `new RegExp(interceptor.requestUrl)` on:
https://github.com/getgauge/taiko/blob/543ad7ab9674ef1adfa0ea4df6a7eec36e62aa03/lib/handlers/networkHandler.js#L66

The result is that Taiko looks for interceptors which match:
`https:\/\/host.net\/api\/path?type=X`
rather than: 
`https:\/\/host\.net\/api\/path\?type=X`
Notice how the `?` and the `.`s in the pattern are not escaped so match with "0 or 1" `h` in the case of the `?`, or "any character" in the case of the `.`.

The only workaround to get our interceptor to match is to create an expression that passes the regex test 
e.g. `".*\\/path\\?type=X"` which means our Testers need to be RegExp fluent
or
e.g. remove the query parameter from the URL: `https://host.net/api/pathtype=X`.

A simple fix for now would be to do a `interceptor.requestUrl === url` check first, then fallback to the regex if that fails.

ed0906:
This fixes #1083 

Currently the `intercept("url", ...)` command does `new RegExp(url)` behind the scenes to match against request URLs. This does not escape special RegExp characters e.g. `.` or `?`

This PR implements a simple String match against the intercept URL & Request URL *and then* does a more complex Reg Exp match if that fails.

This allows us to intercept requests which have query parameters e.g. `https://host.com?query=value` which currently does not work without explicitly escaping the special characters e.g. `https://host\\.com\\?query=value`


ed0906:

zabil:
Fixes #1083

(Adding keyword to trigger automatic closing of the issue on merge)

lib/handlers/networkHandler.js
```diff
@@ -62,9 +62,11 @@ const responseHandler = response => {
 const getMatchingInterceptor = (interceptor, url) => {
   if (interceptor.count === 0) {
     return false;
+  } else if (url === interceptor.requestUrl) {
```
zabil: @ed0906 sorry for the late response, but can we have a test for this in `networkHandler.test.js`?

ed0906: I was going to add one but then I saw how few tests were there already for this class

zabil: It's ok to just add tests around this use case for the PR.

PR #1086, opened 2020-02-11
Chromium download can be skipped when specified in .npmrc

donatasbacius:
Can already be done via an env var but there is no convenient way to skip on every install.

It's useful in constrained environments where one needs to use a specific version of Chromium.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -81,6 +81,14 @@ function main() {
     return;
   }
 
+  if (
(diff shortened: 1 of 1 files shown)


donatasbacius:

sriv:
thanks for the change @donatasbacius

vitolpoc:

donatasbacius:
@sriv - is there anything to do about failing tests? Looks unrelated.

@negiDharmendra - can we push this forward?

sriv:
The test failures are unrelated and transient.

PR #1090, opened 2020-02-17
Automate release

BugDiver:


.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Deploy npm
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          npm publish --access=public
+
+      - name: Bump up version
+        run: |
+          git checkout master && git pull --rebase
+          version=$(npm version patch --no-git-tag-version)
```
sriv: I guess we can add an option to bump `major` or `minor` through some mechanism?

sriv: we can leave it upto the submitter to bump versions in the changeset.

Issue #1095, opened 2020-02-21
[Feature Request] to get an element by indexNumber on elements();


msx752:
hi, could you add a specific function to get element in an array by number of Index

### Case
- there may more than one matches while using

 `$('input[maxlength="11"]');` **or** `$('button[type="button"].btn.btn-outline-dark');`

### Current State
- this is the ugly solution which i found.

```
let matches = $('button[type="button"].btn.btn-outline-dark');
let selectedElement = (await matches.elements())[indexNumber];
```

### Request
- this is not big deal but reducing the line of codes matter while working on project.
```
 let selectedElement =   await `$('input[maxlength="11"]').ElementAt(indexNumber);`
// or
 let selectedElement =   await `$('input[maxlength="11"]').At(indexNumber);`
```
``
### Versions
```
Version: 1.0.3 (Chromium:81.0.3994.0)
```

zabil:
Perhaps

```
let selectedElement =   await $('input[maxlength="11"]').element(indexNumber);
```

PR #1106, opened 2020-02-28
#1095 add support to find element by index

saikrishna321:
Fixes https://github.com/getgauge/taiko/issues/1095

@zabil Can you please review

@msx752

### DIFF
--- a/lib/elementWrapper/elementWrapper.js
+++ b/lib/elementWrapper/elementWrapper.js
@@ -49,6 +49,18 @@ class ElementWrapper {
   async elements(retryInterval, retryTimeout) {
     return await _getIfExists(this._get, this._description)(null, retryInterval, retryTimeout);
   }
+
+  async element(index, retryInterval, retryTimeout) {
+    const results = await _getIfExists(this._get, this._description)(
+      null,
+      retryInterval,
+      retryTimeout,
+    );
+    if (index > results.length - 1) {
+      throw new Error(`Element Index is out of range. Found index range ${results.length - 1}`);
+    }
+    return results[index];
+  }
 }
 
 module.exports = ElementWrapper;
--- a/lib/elementWrapper/elementWrapper.js
+++ b/lib/elementWrapper/elementWrapper.js
@@ -57,7 +57,7 @@ class ElementWrapper {
       retryTimeout,
     );
     if (index > results.length - 1) {
-      throw new Error(`Element Index is out of range. Found index range ${results.length - 1}`);
+      throw new Error(`Element index is out of range. There are only ${results.length} elements`);
     }
     return results[index];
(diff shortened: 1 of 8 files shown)


msx752:
@saikrishna321 it seems good, very useful a function what i need 👍

test/unit-tests/$.test.js
```diff
@@ -117,5 +117,18 @@ describe(test_name, () => {
       const elems = await $('#foo').elements();
       expect(await elems[0].text()).to.be.eql('taiko');
     });
+
+    it('test text() with element index', async () => {
```
zabil: Can we have a test where there are more than one element?
An id is unique so ideally we should be match by something else.

lib/elementWrapper/elementWrapper.js
```diff
@@ -49,6 +49,18 @@ class ElementWrapper {
...
   }
+
+  async element(index, retryInterval, retryTimeout) {
+    const results = await _getIfExists(this._get, this._description)(
+      null,
+      retryInterval,
+      retryTimeout,
+    );
+    if (index > results.length - 1) {
+      throw new Error(`Element Index is out of range. Found index range ${results.length - 1}`);
```
zabil:
```suggestion
      throw new Error(`Element index is out of range. There are only ${results.length} elements`);
```

lib/elementWrapper/elementWrapper.js
```diff
@@ -49,6 +49,18 @@ class ElementWrapper {
...
   }
+
+  async element(index, retryInterval, retryTimeout) {
+    const results = await _getIfExists(this._get, this._description)(
+      null,
+      retryInterval,
+      retryTimeout,
+    );
+    if (index > results.length - 1) {
+      throw new Error(`Element index is out of range. There are only ${results.length} elements`);
```
zabil: One more small change. I noticed that if there are only 1 element the message can read as follows.

```suggestion
      throw new Error(`Element index is out of range. There are only ${results.length} element(s)`);
```

This way the message will be readable for 1 and more than one elements.

test/unit-tests/$.test.js
```diff
@@ -117,5 +117,18 @@ describe(test_name, () => {
...
+      const elems = await $('#foo').element(0);
+      expect(await elems.text()).to.be.eql('taiko');
+    });
+
+    it('Should throw error when element index is out of bound', async () => {
+      await $('#foo')
+        .element(1)
+        .catch(err => {
+          expect(err).to.match(/Element index is out of range. There are only 1 elements/);
```
saikrishna321:
```suggestion
          expect(err).to.match(/Element index is out of range. There are only 1 element(s)/);
```

PR #1115, opened 2020-03-05
Websocket issue

### LINKED ISSUE #896: Error Message: Error: WebSocket is not open: readyState 3 (CLOSED)
### Expected behavior
I have above 40 scenarios using the following functions:

beforeSuite(async () => {
  await openBrowser({ headless: true, args: [
    '--disable-gpu',
    '--disable-dev-shm-usage',
    '--disable-setuid-sandbox',
    '--no-first-run',
    '--no-sandbox',
    '--no-zygote',
    '--window-size=1800,1200'
  ]})})

afterSuite(async () => {
  await closeBrowser()
})

beforeScenario(async () => {
  await deleteCookies()
  await openTab()
})

afterScenario(async () => {
  await closeTab()
})

### Actual behavior
After about 6 scenarios I get the error below:

 Error Message: Error: WebSocket is not open: readyState 3 (CLOSED)
    Stacktrace: 
    Error: WebSocket is not open: readyState 3 (CLOSED)
        at WebSocket.send (node_modules/ws/lib/websocket.js:329:19)
        at Chrome._enqueueCommand (node_modules/chrome-remote-interface/lib/chrome.js:281:18)
        at node_modules/chrome-remote-interface/lib/chrome.js:87:22
        at new Promise (<anonymous>)
        at Chrome.send (node_modules/chrome-remote-interface/lib/chrome.js:86:20)
        at Object.handler [as clearBrowserCookies] (node_modules/chrome-remote-interface/lib/api.js:32:23)
        at module.exports.deleteCookies (node_modules/taiko/lib/taiko.js:669:19)
        at module.exports.<computed> (node_modules/taiko/lib/taiko.js:3542:36)
        at Object.<anonymous> (tests/user.js:26:9)

    Error Message: TypeError: Cannot read property 'message' of undefined
    Stacktrace: 
    TypeError: Cannot read property 'message' of undefined
        at Object.runtimeEvaluate (node_modules/taiko/lib/handlers/runtimeHandler.js:51:18)

### Versions
```
taiko version: 1.0.2 (Chromium: 76.0.3803.0) RELEASE
Gauge version: 1.0.6
Commit Hash: 2bc49db

Plugins
-------
flash (0.0.1)
html-report (4.0.8)
js (2.3.5)
json-report (0.3.2)
screenshot (0.0.1)
spectacle (0.1.3)
xml-report (0.2.2)
```

negiDharmendra:
Fixes #896

### DIFF
--- a/lib/handlers/runtimeHandler.js
+++ b/lib/handlers/runtimeHandler.js
@@ -53,6 +53,9 @@ const runtimeEvaluate = async (exp, executionContextId, opt = {}) => {
   try {
     runtimeResult = await runtime.evaluate(options);
   } catch (e) {
+    if (e.message.match(/Chrome process with pid \d+ exited with/)) {
+      throw e;
+    }
     if (e.response.message === 'Cannot find context with specified id' && executionContextId) {
       executionContextIds = executionContextIds.filter(id => id !== executionContextId);
     }
@@ -85,6 +88,9 @@ const runtimeCallFunctionOn = async (exp, executionContextId, opt = {}) => {
   try {
     runtimeResult = await runtime.callFunctionOn(options);
   } catch (e) {
+    if (e.message.match(/Chrome process with pid \d+ exited with/)) {
+      throw e;
+    }
     if (
       e.response &&
       e.response.message === 'Cannot find context with specified id' &&
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -213,7 +213,7 @@ function hasChromeProcessCrashed() {
 
 async function reconnect() {
   const response = await isReachable(`${currentHost}:${currentPort}`);
-  if (_client && response) {
+  if (response) {
     try {
       logEvent('Reconnecting');
       eventHandler.emit('reconnecting');
@@ -484,7 +484,7 @@ module.exports.switchTo = async arg => {
         'The "targetUrl" argument must be of type string or regex. Received type ' + typeof arg,
       );
     }
-    if (arg.trim() == '') {
+    if (typeof arg === 'string' && arg.trim() === '') {
       throw new Error(
         'Cannot switch to tab or window. Hint: The targetUrl is empty. Please use a valid string or regex',
       );
@@ -662,6 +662,7 @@ module.exports.openTab = async (
 };
 
 /**
+ * @deprecated User openIncognitoWindow
  * Opens the specified URL in the browser's window. Adds `http` protocol to the URL if not present.
  * @example
  * await openWindow('https://google.com', { name: 'windowName' }) - Opens a Incognito window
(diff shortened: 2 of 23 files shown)


lib/taiko.js
```diff
@@ -3177,7 +3206,15 @@ for (const func in module.exports) {
       if (defaultConfig.observe) {
         await waitFor(defaultConfig.observeTime);
       }
-      return await realFuncs[func].apply(this, arguments);
+      try {
+        return await realFuncs[func].apply(this, arguments);
+      } catch (e) {
+        if (e.message.match(/WebSocket is not open: readyState 3/)) {
```
NivedhaSenthil: `WebSocket is not open: readyState 3` may not always mean browser is crashed. It can also happen when we try to do action on closed tab if reconnect fails or after client.close() is called( could happen from taiko plugin  )

negiDharmendra: Agreed we should combine this with some sort of check which will indicate that the `Browser` has crashed. 
> after client.close() is called( could happen from taiko plugin )

We should not allow `plugins` to use `close` method.  I.M.O we should include `close` method in [unsupported client methods list](https://github.com/getgauge/taiko/blob/3c0e34ecf36d90bdb0d329a139ae64e0d0a306e3/lib/taiko.js#L363)

sriv: >after client.close() is called( could happen from taiko plugin ) 

while that may be true, I doubt this is what's happening in the issues reported. There are no reports of any plugin being used.

sriv: But yes, the browser crash is only a guess, and we can be a bit defensive. The other ones are perhaps more easily reproducible.

negiDharmendra:
There are three places we have to handle this issue. 
- At the time when the chrome process crashes or gets killed by some other process (other than Taiko). This can be done by adding a listener for exit event. Here we can make sure that subsequent `openBrowser` call does not result in OpenBrowser cannot be called again as there is a chromium instance open. error.


- Before every call to CDP domain API. Here we can check for connection availability before making the call to the CDP domain.


- If somehow the above two cases are not enough to handle the WebSocket is not open: readyState 3 (CLOSED) error. We can try to handle it in https://github.com/getgauge/taiko/blob/3c0e34ecf36d90bdb0d329a139ae64e0d0a306e3/lib/taiko.js#L3180

lib/taiko.js
negiDharmendra: > But yes, the browser crash is only a guess, and we can be a bit defensive.

The first point mentioned [here](https://github.com/getgauge/taiko/pull/1115#issuecomment-595623034) will insure that Taiko gets notified whenever browser crashes.

NivedhaSenthil:
> * If somehow the above two cases are not enough to handle the WebSocket is not open: readyState 3 (CLOSED) error. We can try to handle it in https://github.com/getgauge/taiko/blob/3c0e34ecf36d90bdb0d329a139ae64e0d0a306e3/lib/taiko.js#L3180

This will be redundant I feel, if the call to domains are handled then this should also be handled as the only way  we make websocket calls are using them

negiDharmendra:
> if the call to domains are handled then this should also be handled as the only way we make websocket calls are using them.

It's not really handeling the `websocket closed`  error. It just do a check before making the call to CDP domain. So, if after that check browser crashes we still need to handle that.

lib/handlers/runtimeHandler.js
```diff
@@ -53,6 +53,9 @@ const runtimeEvaluate = async (exp, executionContextId, opt = {}) => {
   try {
     runtimeResult = await runtime.evaluate(options);
   } catch (e) {
+    if (e.message.match(/Chrome process with pid \d+ exited with/)) {
```
zabil: I am not sure if this will work if we change the browser langauge and internationalisation as we are matching messages in english. 

Can you check this after changing the browser or OS langauge?

negiDharmendra: I tried after changing the langauge to french and every thing works fine.

test/functional-tests/specs/browserActions.spec
```diff
@@ -6,7 +6,7 @@
 * Navigate to "http://localhost:3001/"
 * Click "Multiple Windows"
 * Assert page navigated to "/windows"
-* Click "Click Here"
+* Click "Click Here" with options "{\"navigationTimeout\": 40000,\"waitForNavigation\": true, \"waitForEvents\": [\"firstMeaningfulPaint\", \"DOMContentLoaded\"]}"
```
NivedhaSenthil: we may not need this

negiDharmendra: Make sense after [this](https://github.com/getgauge/taiko/blob/96b369d0533d539ef6556f9d761bad22fc7f0202/lib/doActionAwaitingNavigation.js#L110) change we don't need this

NivedhaSenthil:
If browser or page crashes when the api is executing it just hangs giving no response.

```
> openBrowser()
 ✔ Browser opened
> goto("google.com")
 ✔ Navigated to URL http://google.com
> text("sdfgsfd").exists()
```

close the browser manually after entering the last api.

NivedhaSenthil:
all api hangs after a page crash

`client().Page.crash()` will be better if taiko can report that to user.

lib/taiko.js
```diff
@@ -96,16 +96,42 @@ const createTarget = async () => {
...
+};
 const initCRIProperties = c => {
+  page = createProxyForCDPDomain(c, 'Page');
+  network = createProxyForCDPDomain(c, 'Network');
+  createProxyForCDPDomain(c, 'Runtime');
+  input = createProxyForCDPDomain(c, 'Input');
+  dom = createProxyForCDPDomain(c, 'DOM');
+  overlay = createProxyForCDPDomain(c, 'Overlay');
+  security = createProxyForCDPDomain(c, 'Security');
   _client = c;
```
NivedhaSenthil: Browser,Emulation and Target are missed   from creating proxy

PR #1117, opened 2020-03-06
Support for Color API

### LINKED ISSUE #1032: Add API for color input type
Depends on #612 

For the `input` type `color` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color

```
<p>Choose your monster's colors:</p>

<div>
    <input type="color" id="head" name="head"
           value="#e66465">
    <label for="head">Head</label>
</div>

<div>
    <input type="color" id="body" name="body"
            value="#f6b73c">
    <label for="body">Body</label>
</div>
```

Add the following API

```
color({id: head}).exists()
color({id:head}).select('#2BD3EF')
color(near('body')).value()
```

saikrishna321:
Fixes #1032


BugDiver:
@saikrishna321  Few scenarios which I think can be addressed

*  Using `select` with no parameter throws an exception which is not user friendly
```
Thrown:
Cannot destructure property `stack` of 'undefined' or 'null'.
```
* Using `select` with invalid color code (hashcode or rgb/rgba) returns success but it does not select the color since it's invalid

```
> c.select('asda')
 ✔ Selected asda
> c.select('rgb()')
 ✔ Selected rgb()
> c.select('rgb(asdsd)')
 ✔ Selected rgb(asdsd)
```

saikrishna321:
> * Using `select` with no parameter throws an exception which is not user-friendly

@BugDiver What error message would you recommend?

zabil:
> > * Using `select` with no parameter throws an exception which is not user-friendly
> 
> @BugDiver What error message would you recommend?

`The color code <code> is invalid. Please pass a valid HTML color code.`

BugDiver:
Also, about

> * Using select with invalid color code (hashcode or rgb/rgba) returns success but it does not select the color since it's invalid

I'm not sure if it's possible but if we could validate that the given input is a valid color code that would great.

For Ex: 
```
.select('zzzcsc')
 X The color code 'zzzcsc' is invalid. Please pass a valid HTML color code.
```

PR #1125, opened 2020-03-12
added experimental typescript support

### LINKED ISSUE #209: Add typescript bindings
Add typescript bindings to APIs so that the tests can be written in TypeScript as well.

nuclearglow:
Closes #209 

* updated docs
* exposed typings through package.json
* Bonus: added package.json schema

This solution would mean that everyone using Taiko with TypeScript would be exposed to the Typings provided, but there is also an optional solution which would make the typings optional but requires a bit of configuration work for every project using it:
* the types folder in the repo would contain a taiko folder with a `index.d.ts`
* the typings could be consumed using the "types" field in tsconfig.json
* a user with gauge-ts setup would have to edit `tsconfig.json` in this way:

```js
{
    "compilerOptions": {
        "typeRoots": ["node_modules/@types", "node_modules/taiko/types"],  // add available taiko TypeScript Type Definition folder to the project
        "types": ["node", "taiko"] // use taiko types in the project
    }
}
```

If you prefer to keep this optional, let me know and I will update the PR.
Possible TODOs:
* The typings themselves are workable, but likely need more work to reflect the current API better.
* The API reflected is 1.0.2, I did not include any changes from there to the current version

Hopefully, we can work together to include typings for Taiko :-)


nuclearglow:
Here is an example repo I just created using this PR branch for a Gauge TS Taiko example:
https://github.com/nuclearglow/gauge-ts-taiko-types-example

sriv:
@nuclearglow - what do you suggest is the best way to keep the types in sync? Is there any automated mechanism that you know of, which can be hooked up t0 husky?

Do you think it makes sense to port some tests to TS and make them run in the pipeline?

thanks for the contribution!

PS - i had approved this PR, but then I figured we should discuss these...

nuclearglow:
@sriv yes, I had a look at available tools that produce typings automatically, but my preliminary assessment is that they do not really work with the Taiko API, as it is really flexible and allows a lot in the method parameters, all of this is pretty hard to configure out automatically. So I would recommend that extensive testing of the typings should be implemented. It would certainly mean that a TypeScript project needs to test the Taiko API methods using the available parameters and should check whether the project still compiles. This way, any future changes to the API which breaks the typings would mean that TypeScript tests are failing - and then the typings could be enhanced to reflect the newest API changes. Of course, new methods would need new tests and possibly new typings.

nuclearglow:
I would also recommend that Taiko declares TypeScript support as experimental for the next releases until we work these things out and would therefore only allow projects to set up the typings manually, this way we can get some feedback from real world projects out there if the typings work against their test projects - and at the same time we do not break existing TypeScript test projects where the typings are inadequate. 

If you think that is the way to go, I will update this PR so that the README, the folder structure and the setup only supports manually configured projects, just let me know.

nuclearglow:
The most important thing we should discuss within the Typings themselves is the `TaikoSearchElement` (https://github.com/getgauge/taiko/pull/1125/files#diff-6d9454e03789edeaeb913b3ff375045fR64), as you can see this basically allows everything and it shows a weak point in my typings. But I do not know enough about how the Taiko API works internally to provide stronger typings or a better interface...

sriv:
> So I would recommend that extensive testing of the typings should be implemented.

Agreed.

> I would also recommend that Taiko declares TypeScript support as experimental for the next releases until we work these things out

Fair enough. We usually toggle experimental features off, but manual config to enable it works equally well.

On the types - 

- is there a reason to have `Taiko` prefixed on most (all) types? Would it be possible to use namespaces instead?

Regarding `TaikoSearchElement` -  in taiko, the input to any element search can be either - 
- string 
- instance of elementwrapper
- instance of element

As a result you'll see checks like this: https://github.com/getgauge/taiko/blob/master/lib/taiko.js#L1151

it'd be nice to have types, but we are just touching the interface here. I think it'd be good to constrain `TaikoSeachElement` to just these. Thoughts?

nuclearglow:
Hi @sriv thanks for your feedback, I have worked on the PR and made some updates:
* experimental TypeScript support now needs to be configured manually in the gauge-ts project's `tsconfig.json`, I have updated the README with an example. Also see my updated example repo which is configured to work with the PR branch -> https://github.com/nuclearglow/gauge-ts-taiko-types-example
* the `taiko/index.d.ts` already declare a module `taiko`, so I have removed the `Taiko` prefix everywhere, should work fine, if there are name clashes, usually VS Code or IntelliJ allows to choose from the module, which would then be `taiko` for us. An import in `StepImplementation.ts` would look like this: `import { BrowserOptions } from 'taiko';`
* I have not changed the `TaikoSearchElement` right now. I have taken a deeper look at this and I want to discuss an example from your API: https://docs.taiko.dev/#click. Here, it is possible to provide a string, a selector and an Object, such as {x: int, y: int}. I have not typed this, instead I allow generic objects. Either we provide interfaces to deal with Objects in the taiko API or we use the generic option. I would like the former, but I do not know how much work that would be. What do you think? Should we implement specific interfaces for the `Object` parameters? I might need help to catch them all...

sriv:
:+1:  on (1) and (2).

Regarding `TaikoSearchElement`: I too like the first option of having interfaces rather than generics.

`string` is straightforward. For `selector` and `Object`, taiko looks for `Object` to map to an `Element`. Since JS does not make type safety easy, we added a helper method [`IsSelector`](https://github.com/getgauge/taiko/blob/master/lib/helper.js#L173-L177) and [`IsElement`](https://github.com/getgauge/taiko/blob/master/lib/helper.js#L179).

Does this help? Perhaps we can start porting some of our tests to use the TS bindings?

nuclearglow:
Hey @sriv thanks for the pointers to the helper methods, I managed to learn a bit more. I think we can restrict the SearchElement type to the types you mentioned, but I will need to go through the API and be on the lookout for special cases. I found this in [taiko.js:1300](https://github.com/getgauge/taiko/blob/master/lib/taiko.js#L1300):
```js
 if (isSelector(selector) || isString(selector) || isElement(selector)) {
...
  } else {
    options = setClickOptions(options, selector.x, selector.y);
```

This basically means that there are special cases for the selector which are going further than `selector`, `string` and `element` types. So I propose that we restrict the SearchElement to the instances you mentioned, I will analyse the helper methods to see that I catch everything you check. And for special parameters like in `click()` where a coordinate can be given as a SearchElement, I would then type something like this:

```typescript
selector: SearchElement | Point2D
```
And yes, we could start porting some test cases to use TypeScript, can you help with the initial setup and maybe provide an example if you know to to go about this, then I could help with some tests as far as my project budget allows. If we should think about how to do it, let me know, I will help.

nuclearglow:
OK, a new update with some updated typings after I have taken a look at the helper functions and the API again. Basically, I have upgraded the basic SerachElement definition to allow only the stuff that was also allowed before:
```typescript
    // BasicSelector mimics isSelector
    export interface BasicSelector {
        // FIXME: @sriv: is that correct?
        elements: Element[] | Node[] | string[];
        exists(): boolean;
        [key: string]: any;
    };
    // isSelector also allows ElementWrapper instances
    export type Selector = BasicSelector | ElementWrapper;
    // SearchElement mimics isSelector, isString, isElement
    export type SearchElement = string | Selector | Element;
```

sriv:
Sounds like good progress @nuclearglow ! I will take a closer look soon (feel free to ping me here if I am blocking your work), in the meantime there are some merge conflicts :)

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#reload
+    export function reload(url: string, options?: NavigationOptions): Promise<void>;
+    // https://taiko.gauge.org/#goback
+    export function goBack(options?: NavigationOptions): Promise<void>;
+    // https://taiko.gauge.org/#goforward
+    export function goForward(options?: NavigationOptions): Promise<void>;
+    // https://taiko.gauge.org/#title
+    export function title(): Promise<string>;
+    // https://taiko.gauge.org/#click
+    export function click(selector: SearchElement | MouseCoordinates, options?: ClickOptions, args?: string[]): Promise<void>;
```
NivedhaSenthil, line 238: args is a rest parameter and its type has to be RelativeSearchElement

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#doubleclick
+    export function doubleClick(selector: SearchElement | MouseCoordinates, options?: BasicNavigationOptions, args?: string[]): Promise<void>;
```
NivedhaSenthil, line 240: args is a rest parameter and its type has to be RelativeSearchElement

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#rightclick
+    export function rightClick(selector: SearchElement | MouseCoordinates, options?: BasicNavigationOptions, args?: string[]): Promise<void>;
```
NivedhaSenthil, line 242: args is a rest parameter and its type has to be RelativeSearchElement

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#draganddrop
+    export function dragAndDrop(
+        source: SearchElement,
+        destination: SearchElement,
+        distance: DragAndDropDistance
+    ): Promise<void>;
+    // https://taiko.gauge.org/#hover
+    export function hover(selector: SearchElement, options?: EventOptions): Promise<void>;
```
NivedhaSenthil, line 250: can also take navigationOptions

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#focus
+    export function focus(selector: SearchElement, options?: EventOptions): Promise<void>;
```
NivedhaSenthil, line 252: can also take navigationOptions

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    export interface TapOptions extends BasicNavigationOptions, EventOptions {}
+
+    export interface KeyOptions extends NavigationOptions {
+        text?: string;
+        delay?: number;
+    }
+
+    export interface WriteOptions extends NavigationOptions {
```
NivedhaSenthil, line 59: extends eventOptions too

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+        distance: DragAndDropDistance
+    ): Promise<void>;
+    // https://taiko.gauge.org/#hover
+    export function hover(selector: SearchElement, options?: EventOptions): Promise<void>;
+    // https://taiko.gauge.org/#focus
+    export function focus(selector: SearchElement, options?: EventOptions): Promise<void>;
+    // https://taiko.gauge.org/#write
+    export function write(text: string, into?: SearchElement, options?: WriteOptions): Promise<void>;
+    // https://taiko.gauge.org/#clear
+    export function clear(selector: SearchElement, options?: NavigationOptions): Promise<void>;
```
NivedhaSenthil, line 256: can also take EventsOptions

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+
+    export interface GlobalConfigurationOptions extends BasicNavigationOptions {
+        observeTime?: number;
+        retryInterval?: number;
+        retryTimeout?: number;
+    }
+
+    export interface TapOptions extends BasicNavigationOptions, EventOptions {}
+
+    export interface KeyOptions extends NavigationOptions {
```
NivedhaSenthil, line 54: extends eventOptions too

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#write
+    export function write(text: string, into?: SearchElement, options?: WriteOptions): Promise<void>;
+    // https://taiko.gauge.org/#clear
+    export function clear(selector: SearchElement, options?: NavigationOptions): Promise<void>;
+    // https://taiko.gauge.org/#attach
+    export function attach(filepath: string, to: SearchElement): Promise<void>;
+    // https://taiko.gauge.org/#press
+    export function press(keys: string | string[], options?: KeyOptions): Promise<void>;
+    // https://taiko.gauge.org/#highlight
+    export function highlight(selector: SearchElement): Promise<void>;
```
NivedhaSenthil, line 262: takes ...args rest parameter which is proximityElementSelector[]

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#mouseaction
+    export function mouseAction(action: 'press' | 'move' | 'release', coordinates: MouseCoordinates, options?: NavigationOptions): Promise<void>;
```
NivedhaSenthil, line 264: can also take EventOptions

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#scrollto
+    export function scrollTo(selector: SearchElement, options?: EventOptions): Promise<void>;
```
NivedhaSenthil, line 266: can also take NavigationOptions

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    export function tap(selector: SearchElement, options?: TapOptions, ...args: SearchElement[]): Promise<void>;
+
+    /**
+     * Selectors
+     */
+
+    // https://taiko.gauge.org/#dollar
+    export function $(selector: string, ...args: SearchElement[]): Selector;
+    // https://taiko.gauge.org/#image
+    export function image(selector: SearchElement, ...args: SearchElement[]): SearchElement;
```
NivedhaSenthil, line 287: can take option that says `{selectHiddenElement:true|false}`

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#link
+    export function link(selector: SearchElement, ...args: SearchElement[]): SearchElement;
```
NivedhaSenthil, line 289: can take option that says {selectHiddenElement:true|false}

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#listitem
+    export function listItem(selector: SearchElement, ...args: SearchElement[]): SearchElement;
+    // https://taiko.gauge.org/#button
+    export function button(selector: SearchElement, ...args: SearchElement[]): SearchElement;
```
NivedhaSenthil, line 293: can take option that says {selectHiddenElement:true|false}

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#inputfield
+    export function inputField(selector: SearchElement, ...args: SearchElement[]): ElementWrapper;
```
NivedhaSenthil, line 295: deprecated and removed in latest version

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#filefield
+    export function fileField(selector: SearchElement, ...args: SearchElement[]): ElementWrapper;
```
NivedhaSenthil, line 297: can take option that says {selectHiddenElement:true|false}

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#textbox
+    export function textBox(selector: SearchElement, ...args: SearchElement[]): ElementWrapper;
+    // https://taiko.gauge.org/#combobox
+    export function comboBox(selector: SearchElement, ...args: SearchElement[]): ElementWrapper;
```
NivedhaSenthil, line 301: deprecated and removed in latest version

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#dropdown
+    export function dropDown(selector: SearchElement, ...args: SearchElement[]): ElementWrapper;
```
NivedhaSenthil, line 303: can take option that says {selectHiddenElement:true|false}

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#checkbox
+    export function checkBox(selector: SearchElement, ...args: SearchElement[]): ElementWrapper;
```
NivedhaSenthil, line 305: can take option that says {selectHiddenElement:true|false}

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#radiobutton
+    export function radioButton(selector: SearchElement, ...args: SearchElement[]): ElementWrapper;
```
NivedhaSenthil, line 307: can take option that says {selectHiddenElement:true|false}

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    // https://taiko.gauge.org/#text
+    export function text(selector: string, ...args: SearchElement[]): ElementWrapper;
```
NivedhaSenthil, line 309: can take option that says {exactMatch:true|false}

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+
+    /**
+     * Proximity Selectors
+     */
+
```
NivedhaSenthil, line 314: all proximity selectors returns RelativeSearchElement

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    export function confirm(message: string, callback: Function): void;
+    // https://taiko.gauge.org/#beforeunload
+    export function beforeunload(message: string, callback: Function): void;
+
+    /**
+     * Helpers
+     */
+
+    // https://taiko.gauge.org/#evaluate
+    export function evaluate(
```
NivedhaSenthil, line 342: can take EventOptions too

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+    export function accept(text?: string): Promise<void>;
+    // https://taiko.gauge.org/#dismiss
+    export function dismiss(text?: string): Promise<void>;
+    // https://taiko.gauge.org/#setconfig
+    export function setConfig(options: GlobalConfigurationOptions): void;
+    // https://taiko.gauge.org/#currenturl
+    export function currentURL(): Promise<string>;
+    // https://taiko.gauge.org/#waitfor
+    export function waitFor(time: number): Promise<void>;
+    export function waitFor(element: string, time: number): Promise<void>;
```
NivedhaSenthil, line 365: element can be searchElement

types/taiko/index.d.ts
```diff
@@ -0,0 +1,366 @@
...
+
+    // BasicSelector mimics isSelector
+    export interface BasicSelector {
+        // FIXME: @sriv: is that correct?
+        elements: Element[] | Node[] | string[];
+        exists(): boolean;
+        [key: string]: any;
+    };
+    // isSelector also allows ElementWrapper instances
+    export type Selector = BasicSelector | ElementWrapper;
```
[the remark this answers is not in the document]
nuclearglow: The difference between `Selector` and `SearchElement? in the typings is this: A selector always needs to have the method `exists()` and others to handle it, the SearchElement can be a lot more such as just a string. Am I seeing this wrong? I sometimes have a hard time going through the code that I don't know very well. So I propose the following update:
```typescript
// isSelector also allows ElementWrapper instances
    export type Selector = BasicSelector | ElementWrapper;

    // SearchElement mimics isSelector, isString, isElement and also allows relative searches
    export type SearchElement = string | Selector | Element | RelativeSearchElement;
```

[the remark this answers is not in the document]
nuclearglow: Here is my first try for a `RelativeSearchElement`:
```typescript
    export interface RelativeSearchElement {
        desc: string;
        condition(element: Element, value: number): boolean;
        findProximityElementRects(): { elem: Element, result: any }; // result is wrapped in a callback
        validNodes(nodeId: Element): MatchingNode;
    }
```

types/taiko/index.d.ts
nuclearglow: that should already be the case, since the NavigationOptions extend EventOptions like this:
```typescript
export interface NavigationOptions extends BasicNavigationOptions, EventOptions {}
```

types/taiko/index.d.ts
nuclearglow: changed

types/taiko/index.d.ts
nuclearglow: changed

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: changed

types/taiko/index.d.ts
nuclearglow: changed

nuclearglow:
Hi @NivedhaSenthil thank you for the great feedback, I have tried to add everything as far as I understood it. I still have some trouble with the relative searches, they are hard to type since it is a bunch of options. But I superficially typed the RelativeSearchElement and added it as the return type of the proximity methods, I hope this will work out. At this point, I would need real TS tests to check that everything can be typed as we envision it. I will try to add some to my gauge TS example project soon. Please review my changes and let me know if you have more feedback!

types/taiko/index.d.ts
nuclearglow: changed

types/taiko/index.d.ts
nuclearglow: hmm, 
```typescript
NavigationOptions extends BasicNavigationOptions, EventOptions
and
WriteOptions extends NavigationOptions
```
I think that the extension is transitive and it should already be included...

types/taiko/index.d.ts
nuclearglow: should already be the case

types/taiko/index.d.ts
nuclearglow: should already be the case

types/taiko/index.d.ts
nuclearglow: changed

types/taiko/index.d.ts
nuclearglow: should already be the case

types/taiko/index.d.ts
nuclearglow: changed

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: removed

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: removed

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
nuclearglow: added

types/taiko/index.d.ts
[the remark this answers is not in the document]
NivedhaSenthil: True there are cases where api can take only RelativeSearchElement as selector too. 
Eg: `textBox(near('something'))`. This looks good :+1:

NivedhaSenthil:
PR looks fine for an initial TS support, this is a great effort :) Will try adding some test in TS and port some existing test also.. Do follow https://github.com/getgauge/taiko/issues/1152 for any progress.

klctrl:
Hi there. If I understand this thread correctly then the experimental Typescript is only available if Taiko is used in conjunction with Gauge. Is this correct? If not, could someone maybe explain (to a non-expert) how Typescript could be used for writing test cases if other test frameworks (e.g. Jest, Mocha) are used?

zabil:
> Typescript is only available if Taiko is used in conjunction with Gauge.

The typescript bindings are for Taiko. You can use these binding with any test runner. You can refer https://docs.taiko.dev/experimental_features/#typescript-support on adding configuration to [`tsconfig.json`](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html)

klctrl:
@zabil Great, that has clarified it, thanks a lot!

Issue #1031, opened 2020-01-14
Add API for the range input type


zabil:
Depends on #612

Add an API for working with the range input type for example

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

```
<div>
  <input type="range" id="volume" name="volume"
         min="0" max="11">
  <label for="volume">Volume</label>
</div>
```

```
range({id: 'volume'}).exists()
range({id: 'volume'}).select(10)
range({min: 0, max: 11}).exists()
range({min: 0, max: 11}).select(11)
range(near("Volume")).select(10)
range({min: 0, max: 11}).value()
```

Geeky info https://youtu.be/uMSV4OteqBE?t=78

o-x-y-g-e-n:
I will separate this from `textBox` API.

NivedhaSenthil:
Implemented via 6e8e23c

PR #1129, opened 2020-03-16
#1031 Added support for range API

saikrishna321:
Co-authored-by: priyanupur <priya.seiji@gmail.com>


BugDiver:
@saikrishna321 Few scenarios to take care

* When `.select` is used with a number higher than max, The output is
```
The value 30 should be between the minimum range  or maximum range of 20
 ✔ Selected 30
```

But it selects 20. IMO the message to tell that it has selected 20.

* When `.select` is used with a number lower than the minimum, The output is

```
The value 1 should be between the minimum range 10 or maximum range of 20
 ✔ Selected 1
```

But it selects 10. IMO the message to tell that it has selected 10. 

* When `.select` is used with a decimal number, The output is
```
✔ Selected 10.2
```
But `.value()` returns `10`, since it could not select decimal values.
IMO the decimal input should be allowed only when `step` attribute is defined in `input` tag and it is a decimal value. Also if the value of step is not divisor if the value to be selected It should inform. (print the value which actually got selected, the closest one )

@zabil Thoughts?

priyamish:
> @saikrishna321 Few scenarios to take care
> 
> * When `.select` is used with a number higher than max, The output is
> 
> ```
> The value 30 should be between the minimum range  or maximum range of 20
>  ✔ Selected 30
> ```
> 
> But it selects 20. IMO the message to tell that it has selected 20.
> 
> * When `.select` is used with a number lower than the minimum, The output is
> 
> ```
> The value 1 should be between the minimum range 10 or maximum range of 20
>  ✔ Selected 1
> ```
> 
> But it selects 10. IMO the message to tell that it has selected 10.
> 
> * When `.select` is used with a decimal number, The output is
> 
> ```
> ✔ Selected 10.2
> ```
> 
> But `.value()` returns `10`, since it could not select decimal values.
> IMO the decimal input should be allowed only when `step` attribute is defined in `input` tag and it is a decimal value. Also if the value of step is not divisor if the value to be selected It should inform. (print the value which actually got selected, the closest one )
> 
> @zabil Thoughts?

Thanks for the feedback, @BugDiver :
1&2. Fixing the messages makes sense, we will make it more intuitive for the consumers.
3. @saikrishna321  and I had a conversation regarding this. We will take a look at this one, to see how can we incorporate this.

saikrishna321:
@BugDiver @zabil Please review. Fixed comments

test/unit-tests/range.test.js
```diff
@@ -0,0 +1,62 @@
...
+    );
+  });
+
+  it('Should throw warning when exceeded maximum value', async () => {
+    sinon.stub(console, 'warn');
+    await range({ id: 'range-1' }).select(110);
+    expect(console.warn.calledOnce).to.be.true;
+    expect(
+      console.warn.calledWith(
+        'The value 110 should be between the minimum range 0 or maximum range 100',
```
zabil:
```suggestion
        'The value 110 is not between the input's range of 0-100',
```

test/unit-tests/range.test.js
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  it('Set Range value with Float value', async () => {
+    await range({ id: 'range-1' }).select(10.81);
+    expect(await range({ id: 'range-1' }).value()).to.be.equal('11');
+  });
+
+  it('Should throw error when value is non integer', async () => {
+    await expect(range({ id: 'range-1' }).select('10')).to.be.rejectedWith(
+      'The range value should be Int or Float. Please pass a valid value.',
```
zabil: Please convert this to an Int or Float and then try it before validating.
This is because we don't want to enforce type safety in a langauge that is not type safe and we can do our best to convert it.
In the sample above '10' appears to be a valid value.
If the value is something like 'xy' then let's show the generic error.

```suggestion
      'The value xy is not between the input's range of 0-100',
```

lib/elements/range.js
```diff
@@ -0,0 +1,58 @@
...
+        arg: value,
+        returnByValue: true,
+      }));
+      if (value < result.value.min || value > result.value.max) {
+        console.warn(
+          `The value ${value} should be between the minimum range ${result.value.min} or maximum range ${result.value.max}`,
+        );
+      }
+    });
+    descEvent.emit('success', `Given value is ${value} Selected value is ${result.value.current}`);
```
zabil:
```suggestion
    descEvent.emit('success', `Selected value ${result.value.current} for the given input value ${value}`);
```

I don't see a test for this.

saikrishna321:
@zabil Please review now

BugDiver:
@saikrishna321 I guess the case is not handled when there are not `min` and `max` values.

for a simple `<input type="range" name="range" id="range">`
I get 
```
> r.value()
 value: '50'
> r.select(10)
The value 10 is not between the input's range of -
 ✔ Selected value 10 for the given input value 10
>
```

Notice the warning message.

PR #1136, opened 2020-03-17
Fix waiting for loadEventFired

NivedhaSenthil:
- since in cases like new tab open `loadEventFired` can be missed before taiko makes a new connection wait for page's readyState to be complete rather

### DIFF
--- a/lib/doActionAwaitingNavigation.js
+++ b/lib/doActionAwaitingNavigation.js
@@ -1,6 +1,7 @@
-const { timeouts, wait } = require('./helper');
+const { timeouts, wait, waitUntil } = require('./helper');
 const networkHandler = require('./handlers/networkHandler');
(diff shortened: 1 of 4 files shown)


lib/doActionAwaitingNavigation.js
```diff
@@ -114,12 +103,26 @@ const waitForPromises = (promises, waitForStart) => {
 const waitForNavigation = (timeout, promises = []) => {
   return new Promise((resolve, reject) => {
     Promise.all(promises)
-      .then(resolve)
+      .then(() => {
+        waitUntil(
+          async () => {
+            return (
+              (await runtimeHandler.runtimeEvaluate('document.readyState')).result.value ===
```
saikrishna321: giphy (2)

saikrishna321: Nice catch 🐞

PR #1151, opened 2020-03-24
Fix switchTo to accept regex

### LINKED ISSUE #1147: switchTo() doesn't support RegEx's anymore
There seems to be a bug in `lib/taiko.js` on line 418-422. This function isn't present on a RegEx. The call could be moved. The tests should probably also be updated to test for this use case.

Alternatively this should be removed from the docs.

### Expected behavior
Calling `switchTo(/http(s):\/\/www.google.com/)` is a valid use case.

### Actual behavior
A TypeError is thrown:
```
Error Message: TypeError: arg.trim is not a function
        Stacktrace: 
        TypeError: arg.trim is not a function
            at module.exports.switchTo (node_modules/taiko/lib/taiko.js:419:13)
            at module.exports.<computed> (node_modules/taiko/lib/taiko.js:3225:36)
            at Object.<anonymous> (tests/login.js:70:9)
```

### Steps to reproduce
1. Create a taiko step
2. Call switchTo(/http/)

### Versions
```
$ node --version
v13.11.0

$ taiko --version
Version: 1.0.5 (Chromium: 81.0.3994.0) RELEASE
```

NivedhaSenthil:
- update empty string check
- add tests for regex switchTo
- update wait for readystate to timeout on navigation timeout instead of default.

Fixes #1147

### DIFF
--- a/lib/doActionAwaitingNavigation.js
+++ b/lib/doActionAwaitingNavigation.js
@@ -112,10 +112,10 @@ const waitForNavigation = (timeout, promises = []) => {
             );
           },
           defaultConfig.retryInterval,
-          defaultConfig.retryTimeout,
+          timeout,
         )
           .then(resolve)
-          .catch(reject);
+          .catch(() => reject('Timedout'));
       })
       .catch(reject);
     const timeoutId = setTimeout(() => reject('Timedout'), timeout);
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -422,7 +422,7 @@ module.exports.switchTo = async arg => {
         'The "targetUrl" argument must be of type string or regex. Received type ' + typeof arg,
       );
     }
-    if (arg.trim() == '') {
+    if (typeof arg === 'string' && arg.trim() === '') {
       throw new Error(
         'Cannot switch to tab or window. Hint: The targetUrl is empty. Please use a valid string or regex',
       );
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "taiko",
-	"version": "1.0.7",
+	"version": "1.0.6",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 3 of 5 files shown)


BugDiver:
> @NivedhaSenthil Thank you for contributing to taiko. Your pull request has been labeled as a release candidate 🎉🎉.
> 
> Merging this PR will trigger a release.
> 
> ### Please bump up the version as part of this PR.
> Instructions to bump the version can found at [CONTRIBUTING.md](https://github.com/getgauge/taiko/blob/master/CONTRIBUTING.md)
> 
> If the CONTRIBUTING.md file does not exist or does not include instructions about bumping up the version, please looks previous commits in git history to see what changes need to be done.

This won't be needed. The verison is alread updated.

lib/doActionAwaitingNavigation.js
```diff
@@ -112,10 +112,10 @@ const waitForNavigation = (timeout, promises = []) => {
             );
           },
           defaultConfig.retryInterval,
-          defaultConfig.retryTimeout,
+          timeout,
         )
           .then(resolve)
-          .catch(reject);
+          .catch(() => reject('Timedout'));
```
sriv: Won't this cause all errors to be rejected with `Timedout`?

lib/doActionAwaitingNavigation.js
```diff
@@ -112,10 +112,10 @@ const waitForNavigation = (timeout, promises = []) => {
             );
           },
           defaultConfig.retryInterval,
-          defaultConfig.retryTimeout,
+          timeout,
```
sriv: why should this be changed? `timeout` is used below to reject when timer expires.

lib/doActionAwaitingNavigation.js
NivedhaSenthil: only waiting for readyState to be complete will get rejected with Timedout

lib/doActionAwaitingNavigation.js
NivedhaSenthil: To honour the user given navigation timeout.. if default is used this can timeout before the user given time and the promise will get rejected.

PR #1156, opened 2020-03-26
Add options doc string for text API

saikrishna321:
Fixes https://github.com/getgauge/taiko/issues/766

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -771,7 +771,7 @@ module.exports.closeWindow = async arg => {
  * Closes the specified browser window.
  * @example
  * await closeIncognitoWindow('windowName') - Close incognito window
- * @param {string}  windowName - incognito window name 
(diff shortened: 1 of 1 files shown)


lib/taiko.js
```diff
@@ -2762,8 +2762,9 @@ module.exports.radioButton = (attrValuePairs, _options, ...args) => {
  * await highlight(text('Vehicle'))
  * await text('Vehicle').exists()
  * await text('Vehicle', below('text')).exists()
- *
+ * await text('Vehicle', { exactMatch: true }, below('text')).exists()
  * @param {string} text - Text to match.
+ * @param {Object} options - { exactMatch: true } - Look for exact match
```
zabil:
```suggestion
 * @param {boolean} [options.exactMatch=true] - Look for exact match, eg. { exactMatch: true | false }
```

Issue #1166, opened 2020-03-31
closeBrowser fails


nathanBrenner:
### Expected behavior
There shouldn't be a type error and the browser should close when `closeBrowser` is invoked in the callback of the afterSuite

### Actual behavior
Chromium stays open. Error printed in console:
```
Error Message: TypeError: Cannot read property 'removeAllListeners' of null
  Stacktrace: 
  TypeError: Cannot read property 'removeAllListeners' of null
      at _closeBrowser (node_modules/taiko/lib/taiko.js:347:19)
```
### Steps to reproduce
1. click "Run Scenario
2. Chromium opens, and the error is logged to the console, and chromium stays open

test:
```
## Search

* Starting from a new session
* I visit the console homepage
* I search for vulnerabilities in the ecosystem, version and the results include the vulnerability
  | ecosystem | version | foo  |
  | --------- | ------- |----------------|
  | value1      | version1   | foo1  |
```

### Versions
```
node: 10.16.0, taiko: 1.0.6
```

nathanBrenner:
Works as expected with taiko installed at version 1.0.2. on 1.0.0 and 1.0.3, this error is returned:
```
  Error Message: Error: Timed out
  Stacktrace: 
  Error: Timed out
      at Timeout.<anonymous> (/Users/myworkspace/.gauge/plugins/js/2.3.8/src/test.js:43:23)
      at ontimeout (timers.js:436:11)
      at tryOnTimeout (timers.js:300:5)
      at listOnTimeout (timers.js:263:5)
      at Timer.processTimers (timers.js:223:10)
```
and the original error comes back in 1.0.4, then the timeout bug comes back in 1.0.5

nathanBrenner:
The spec in the description wasn't helpful. I had two files in the tests directory with these lines in both files:

```
beforeSuite(async () => {
  await openBrowser({ headless: headless })
})

afterSuite(async () => {
  await closeBrowser()
})
```

The outcome was openBrowser was called twice, so opening two instances of Chromium, and closeBrowser was called twice, so maybe _client on the second call didn't have the `removeAllListeners`

It also looks like the order of specs that are invoked are sorted alphabetically by the specification headings, and whichever is the first heading, that one and only that one should openBrowser and closeBrowser in the beforeSuite and afterSuite.

is a suite a test specification? or does it comprises of all of the specs in the specs directory?

zabil:
[...]

PR #1181, opened 2020-04-06
Fix close browser

### LINKED ISSUE #818: Taiko should fail if user tries to open or close browser multiple times.
**Actual Behaviour**
If openBrowser is called multiple times, multiple chromium instances are launched. Taiko retains the references and the CRI connection only to the most recent chromium instance.

If close Browser is called multiple times, Taiko is able to close only the most recent chromium instance. Since, the references to other chromium instances are lost, it fails to close them.
```
TypeError: Cannot read property ‘removeAllListeners’ of null
       at Chrome.reconnect (...../...../....../node_modules/taiko/lib/taiko.js:137:13)
       at Chrome.emit (events.js:198:13)
       at WebSocket._ws.on (...../...../....../node_modules/chrome-remote-interface/lib/chrome.js:233:22)
       at WebSocket.emit (events.js:198:13)
       at WebSocket.emitClose (...../...../....../node_modules/ws/lib/websocket.js:191:10)
       at Socket.socketOnClose (...../...../....../node_modules/ws/lib/websocket.js:850:15)
       at Socket.emit (events.js:198:13)
       at TCP._handle.close (net.js:606:12)
```
**Expected Behaviour**
Taiko should throw an error to user mentioning that openBrowser cannot be called again as there is a chromium instance open(closeBrowser is not yet called).

Taiko should fail with appropriate message, if close browser is called multiple times without calling openBrowser.

NivedhaSenthil:
Fixes #818 #1166

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -3344,10 +3344,12 @@ const validate = () => {
   if (_client._ws.readyState > 1) {
     if (chromeProcess && chromeProcess.killed) {
       throw new Error(
-        'Browser is closed due to `closeBrowser()` or crash. Try reopening by `openBrowser()`',
+        'The Browser instance was closed either via `closeBrowser()` call, or it crashed for reasons unknown to Taiko. You can try launching a fresh instance using `openBrowser()` or inspect the logs for details of the possible crash.',
       );
     }
-    throw new Error('Connection to browser lost');
+    throw new Error(
+      "Connection to browser lost. This probably isn't a problem with Taiko, inspect logs for possible causes.",
+    );
   }
 };
 
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -3342,7 +3342,12 @@ const validate = () => {
     throw new Error('Browser or page not initialized. Call `openBrowser()` before using this API');
   }
   if (_client._ws.readyState > 1) {
-    throw new Error('Connection to Browser closed. Try reconnecting by `openBrowser()`');
+    if (chromeProcess && chromeProcess.killed) {
+      throw new Error(
+        'Browser is closed due to `closeBrowser()` or crash. Try reopening by `openBrowser()`',
+      );
+    }
+    throw new Error('Connection to browser lost');
   }
 };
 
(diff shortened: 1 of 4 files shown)


lib/taiko.js
```diff
@@ -3338,9 +3338,12 @@ const evaluate = async (selector, func) => {
 };
 
 const validate = () => {
-  if (!dom || !page) {
+  if (!_client) {
     throw new Error('Browser or page not initialized. Call `openBrowser()` before using this API');
   }
+  if (_client._ws.readyState > 1) {
+    throw new Error('Connection to Browser closed. Try reconnecting by `openBrowser()`');
```
sriv, line 3345: If browser instance is already running, `openBrowser` will error out stating only one instance is allowed. If there isn't a browser instance, it will launch a new one, but we cannot consider that as 'reconnecting'

NivedhaSenthil: Updated the messaging to reflect if the browser instance is closed.

lib/taiko.js
```diff
@@ -3338,9 +3338,17 @@ const evaluate = async (selector, func) => {
...
 
 const validate = () => {
-  if (!dom || !page) {
+  if (!_client) {
     throw new Error('Browser or page not initialized. Call `openBrowser()` before using this API');
   }
+  if (_client._ws.readyState > 1) {
+    if (chromeProcess && chromeProcess.killed) {
+      throw new Error(
+        'Browser is closed due to `closeBrowser()` or crash. Try reopening by `openBrowser()`',
```
sriv, line 3347: Suggestion:

`The Browser instance was closed either via 'closeBrowser()' call, or it crashed for reasons unknown to Taiko. You can try launching a fresh instance using 'openBrowser()' or inspect the logs for details of the possible crash.`

lib/taiko.js
```diff
@@ -3338,9 +3338,17 @@ const evaluate = async (selector, func) => {
...
+      );
+    }
+    throw new Error('Connection to browser lost');
```
sriv, line 3350: Suggestion:

`Connection to browser lost. This probably isn't a problem with Taiko, inspect logs for possible causes.`

PR #1187, opened 2020-04-09
fix element screenshot with proximity selectors

saikrishna321:
Address #1186


lib/handlers/domHandler.js
```diff
@@ -63,7 +64,7 @@ async function assertBoundingBox(e) {
 async function getBoxModel(e) {
   return await dom
     .getBoxModel({
-      nodeId: e,
+      nodeId: isObject(e) ? e.nodeId : e,
```
negiDharmendra, line 67: Can we use `isElement` instead of `isObject`.
`nodeId: isElement(e) ? e.get() : e,`

PR #1193, opened 2020-04-13
send nodeId when capturing screenshot of an element

sriv:


sriv:
too many conflicts- will try to clean and raise a new one.

PR #1194, opened 2020-04-14
fix screenshot for specific element with index

saikrishna321:

### DIFF
--- a/lib/handlers/pageHandler.js
+++ b/lib/handlers/pageHandler.js
@@ -1,8 +1,8 @@
 const { createJsDialogEventName } = require('../util');
-const { handleUrlRedirection, isElement } = require('../helper');
+const { handleUrlRedirection, isElement, isSelector } = require('../helper');
 const { eventHandler } = require('../eventBus');
 const { logEvent } = require('../logger');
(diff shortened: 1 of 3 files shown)


lib/handlers/pageHandler.js
```diff
@@ -1,5 +1,5 @@
 const { createJsDialogEventName } = require('../util');
-const { handleUrlRedirection } = require('../helper');
+const { handleUrlRedirection, isElement } = require('../helper');
 const { eventHandler } = require('../eventBus');
 const { logEvent } = require('../logger');
 const { isSelector, findElements } = require('../elementSearch');
```
negiDharmendra, line 5: Can we use the `isSelector` from [helper.js](https://github.com/getgauge/taiko/blob/374ad9d7ac1c987e2d1c7c1cb98426ce44e06b98/lib/helper.js#L207). 
Please remove the re-export of `isSelector` from [elementSearch.js](https://github.com/getgauge/taiko/blob/374ad9d7ac1c987e2d1c7c1cb98426ce44e06b98/lib/elementSearch.js#L325)

negiDharmendra:
Fixes #1186

PR #1222, opened 2020-04-28
Refactor unit test

NivedhaSenthil:


test/unit-tests/isDisabled.test.js
```diff
@@ -13,134 +13,72 @@ let {
 } = require('../../lib/taiko');
 let { createHtml, removeFile, openBrowserArgs, resetConfig } = require('./test-util');
 let test_name = 'tsesr';
```
sriv, line 15: I don't understand this string and it's usage in the next line

test/unit-tests/openBrowser.test.js
```diff
@@ -6,23 +6,6 @@ let { openBrowser, closeBrowser, client } = taiko;
 
 let { openBrowserArgs } = require('./test-util');
 describe('OpenBrowser', () => {
-  describe('opens browser successfully', () => {
```
sriv, line 9: why are these tests deleted?

NivedhaSenthil: since the tests were not testing anything concrete ... one was ignored and the other one was asserting on a taiko function `client` which will never be null

test/unit-tests/isDisabled.test.js
NivedhaSenthil: seems a typo.. will push a fix

NivedhaSenthil: done 👍

Issue #1226, opened 2020-04-29
Possible Memory Leak?


Serrvosky:
Hi everyone,

After some time executing taiko, it closes with this error and I don't know why. Can you please check the output below?

```
<--- Last few GCs --->

[22973:0x40204c0] 13388428 ms: Scavenge 2029.4 (2045.7) -> 2024.7 (2045.7) MB, 6.3 / 0.0 ms  (average mu = 0.235, current mu = 0.232) allocation failure 
[22973:0x40204c0] 13388478 ms: Scavenge 2029.7 (2045.7) -> 2025.2 (2046.4) MB, 7.1 / 0.1 ms  (average mu = 0.235, current mu = 0.232) allocation failure 
[22973:0x40204c0] 13388532 ms: Scavenge 2030.1 (2046.4) -> 2025.6 (2057.7) MB, 9.3 / 0.1 ms  (average mu = 0.235, current mu = 0.232) allocation failure 


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0xa02f90 node::Abort() [node]
 2: 0xa033b5 node::OnFatalError(char const*, char const*) [node]
 3: 0xb76ffe v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 4: 0xb77379 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 5: 0xd23ad5  [node]
 6: 0xd24166 v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [node]
 7: 0xd309e5 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [node]
 8: 0xd31895 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 9: 0xd320aa v8::internal::Heap::FinalizeIncrementalMarkingIfComplete(v8::internal::GarbageCollectionReason) [node]
10: 0xd3595f v8::internal::IncrementalMarkingJob::Task::RunInternal() [node]
11: 0xc76983 v8::internal::CancelableTask::Run() [node]
[...]

PR #1242, opened 2020-05-08
#1226 fix potential memory leak issue

NivedhaSenthil:

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -85,7 +85,7 @@ let chromeProcess,
 
 module.exports.emitter = descEvent;
 
-const createTarget = async () => {
+const createTarget = async (n) => {
   try {
     const browserTargets = await cri.List({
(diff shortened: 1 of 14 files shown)


lib/taiko.js
```diff
@@ -61,7 +61,7 @@ const path = require('path');
 const { eventHandler } = require('./eventBus');
 const overlayHandler = require('./handlers/overlayHandler');
 const { highlightElement } = require('./elements/elementHelper');
-const numRetries = process.env.TAIKO_CRI_CONNECTION_RETRIES || 10;
+const numRetries = process.env.TAIKO_CRI_CONNECTION_RETRIES || 50;
```
negiDharmendra, line 64: I.M.O it should be inside the [config](https://github.com/getgauge/taiko/blob/master/lib/config.js).

PR #1278, opened 2020-05-20
Reference type declarations in package.json

sceee:
It makes type definitions to be correctly resolved by IDEs. Without, taiko has no discoverable typedefs.

See https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package for further explanation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "bin": {
     "taiko": "bin/taiko.js"
   },
+  "types": "./types/taiko/index.d.ts",
   "scripts": {
     "lint": "eslint --fix . --ext ts,tsx .",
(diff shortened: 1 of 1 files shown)


sceee:
It has nothing to do with this PR but I noticed that the CLA link here https://github.com/getgauge/taiko/blob/master/CONTRIBUTING.md#how-do-i-start-contributing is broken so I have no chance to sign it:
image

zabil:
> I noticed that the CLA link here is broken

Taiko now uses DCO's instead of CLA. And it looks like you've signed your commits which is good enough. I've updated the contribution instructions to reflect that. Thanks for pointing that out.

Issue #1280, opened 2020-05-20
Unable to select slider and move it


amitsarkar007:
### Expected behavior
Should be able to select a slider on a page and interact with it using XPath selectors or CSS selectors in the absence of an ID or a CLASS

### Actual behavior
Unable to interact with slider on the page [Horizontal Slider](https://the-internet.herokuapp.com/horizontal_slider) using the **range** selector. Tried all the following options but still not able to move the slider.

```
range({'class':'sliderContainer'}).select('3.5')
range($(`[type=range]`)).select(3.5)
range($(`[type=range]`)).exists()
range({class:'sliderContainer'}).select(3.5)
range({'class':'sliderContainer'}).select('3.5')
range({id:'range'}).select('3.5')
range({id:'range'}).select(3.5)
range($(`.sliderContainer`)).select(3.5)
```

### Steps to reproduce
1. Go to the page
2. Identify the slider element
3. Interact with it and try to move the slider

### Versions
```
Taiko 
1.0.11 (84.0.4143.0)

Node
v12.16.3
```

NivedhaSenthil:
Thanks for reporting, I could see the value is getting updated in the element but not reflected in DOM as the change events are not triggered properly. We will work on fixing this

NivedhaSenthil:
Fixed by #1299

amitsarkar007:
The [commit](https://github.com/getgauge/taiko/commit/c75497243c46c77612f4b067ba9b16b70b549731) has a test that indicates the issue has been tested and fixed.

But using the code below indicates it is not working
```
taiko

Version: 1.0.12 (Chromium:81.0.3994.0)
Type .api for help and .exit to quit

> openBrowser()
[PASS] Browser opened
> goto("https://the-internet.herokuapp.com/horizontal_slider")
[PASS] Navigated to URL https://the-internet.herokuapp.com/horizontal_slider
> range({id:'range'}).select(3.5)
> Uncaught Range Type[@id = concat('range', "")] not found
> highlight($(`//input`))
[PASS] Highlighted the custom selector $(//input)
> clearHighlights()
[PASS] Cleared all highlights for this page.
> range($(`//input`))
RangeWrapper {
  _get: [AsyncFunction: get],
  _description: 'Range Type with label [object Object] '
}
> highlight($(`[id=range]`))
[PASS] Highlighted the custom selector $([id=range])
> range($(`[id=range]`)).exists
[AsyncFunction: exists]
[...]

PR #1299, opened 2020-06-03
Move the slider and pass the currentvalue to slider

Manikkumar1988:
#1280


sriv:
Do not merge untill https://github.com/getgauge/taiko/pull/1311 is merged. Both can go in the same release.

Issue #301, opened 2018-12-21
Spike on support for firefox


NivedhaSenthil:
Some references,
Juggler (patch on firefox to support automation protocol) - https://github.com/GoogleChrome/puppeteer/tree/master/experimental/juggler
Juggler firefox builds - https://storage.googleapis.com/juggler-builds/

NivedhaSenthil:
Currently [Juggler's protocol](https://github.com/GoogleChrome/puppeteer/blob/master/experimental/juggler/src/Protocol.js) has some variations from CDP, thus taiko may not work out of box with Juggler builds.  

If we have to support we will have to reimplement taiko apis using Juggler protocol.
Some reference from firefox regarding CDP experiments https://wiki.mozilla.org/WebDriver/Meetings/2018-11-26.

RBLG:
Hi,
i am a bit confused by the current state of taiko support for firefox, is there some news on it since 2018?
from what i understand so far, taiko need CDP-like API in firefox and since 2018 it seem like firefox made that CDP API under the name Remote (at least as experimental)?
https://wiki.mozilla.org/Remote
So, does that mean anything for taiko, can we expect firefox to be suported in a near future?
(i would like to help but i dont really know how soooo :/ )

sriv:
Hi @rblg, You are correct, Firefox has support for CDP, but not the entire protocol. I believe, Firefox plans to implement CDP endpoints required to support puppeteer for firefox. Taiko requires additional CDP message/events. So the last I checked Taiko was only able to bring up Firefox, but navigation etc are broken.

RBLG:
@sriv ok i see,
is there still hope for a firefox support from what we got/know so far or more waiting has to be expected for anything to be doable?

RBLG:
im asking because if it is possible i could propose to my company to help implement it since we would require it

sriv:
I see where you're coming from, but there is not much that we can do in Taiko to enable this. I believe firefox remote team is working on CDP support, so I wouldn't rule this out just yet. However unless there is support from browser, there is no guarantee.

RBLG:
Ok, thank you for the answers!

mrpink76:
Just read that puppeteer 2.1.0 now has firefox support 🙂

zabil:
Done via https://github.com/getgauge/taiko/pull/1317#event-3452941757

PR #1300, opened 2020-06-04
Use object id instead of node id

NivedhaSenthil:
- replace nodeid with objectid removing conversion
- cleanup execution context on every new session created
- remove deprecated openwindow and closeWindow apis
- cleanup tests using rewire (require them inside blocks to avoid affecting globals and remove listeners after testing)

First step towards #301


test/unit-tests/browserContext.test.js
```diff
@@ -1,162 +0,0 @@
-const expect = require('chai').expect;
```
sriv, line 1: just want to confirm that this test can be deleted?

test/unit-tests/handlers/targetHandler.test.js
```diff
@@ -14,6 +16,17 @@ describe('TargetHandler', () => {
       targetHandler.__set__('cri', mockCri);
     });
 
+    after(() => {
+      const createdSessionListener = targetHandler.__get__('createdSessionListener');
+      targetHandler
+        .__get__('eventHandler')
+        .removeListener('createdSession', createdSessionListener);
+      targetHandler = rewire('../../../lib/handlers/targetHandler');
+      targetHandler
```
sriv, line 25: why do this in `after`? I see this in other files too. I mean, removing listener on the mock?

test/unit-tests/intercept.test.js
```diff
@@ -1,161 +0,0 @@
-const expect = require('chai').expect;
```
sriv, line 1: again, making sure that this deletion is intentional

test/unit-tests/openTab.test.js
```diff
@@ -26,6 +26,10 @@ describe('openTab', () => {
     taiko.__set__('_client', new EventEmitter());
   });
 
+  after(() => {
+    taiko = rewire('../../lib/taiko');
```
sriv, line 30: this is done on the `before` - why do it in `after`?

test/unit-tests/browserContext.test.js
NivedhaSenthil: yes as the deprecated api is now removed `openWindow` and `closeWindow` removed its corresponding test file too

test/unit-tests/handlers/targetHandler.test.js
NivedhaSenthil: unlike require each time rewire is called it creates a new instance and attaches the listener too... hence gotta remove them so that it does not affect other tests
https://github.com/jhnns/rewire#caveats

test/unit-tests/intercept.test.js
NivedhaSenthil: this was not intentional.. removed accidentally will be adding it back

test/unit-tests/openTab.test.js
NivedhaSenthil: rewire does not have a reset option available... calling that again resets the updated values https://github.com/jhnns/rewire/issues/71

Issue #1259, opened 2020-05-13
closeTab(URL) doesn't exact match the URL


scottdavis99:
### Expected behavior
closeTab(URL) matches the URL exactly, or is able to differentiate between different tabs whose URL starts the same but the full path is unique. 

### Actual behavior
closeTab(URL) doesn't exactly match the URL. If I have two tabs open, one with 'https://thirstyhead.com/conferenceworks' and the other with 'https://thirstyhead.com/groceryworks', the action closeTab('https://thirstyhead.com/groceryworks') closes both tabs. 

### Steps to reproduce
openBrowser()
goto('https://thirstyhead.com/conferenceworks/')
openTab('https://thirstyhead.com/groceryworks/')
switchTo('ConferenceWorks')
closeTab('https://thirstyhead.com/groceryworks/')

[RESULT: closes both tabs and browser]

### Versions
```
Version: 1.0.9 (Chromium:83.0.4103.0)
```

Manikkumar1988:
Issue is tied up with #626 

Need a better way to compare URLS without host & protocol and taking host redirection into the account.

NivedhaSenthil:
Fixed by #1311

PR #1311, opened 2020-06-11
fix for close tab with multi path issue

Manikkumar1988:
#1259


lib/handlers/targetHandler.js
```diff
@@ -40,18 +40,31 @@ const isMatchingRegex = function (target, targetRegex) {
...
+	}
+	return targetUrl;
+};
+
 const isMatchingUrl = function (target, targetUrl) {
   if (targetUrl instanceof RegExp) {
     return false;
   }
+
+  targetUrl = prependHttp(targetUrl);
```
sriv, line 56: `prependHttp` is parsing the url again, while the url is getting parsed in line 58 in the calling method.

I wonder if line 61 can simply return `http` as the default host?

Manikkumar1988:
```
const url = require('url');
console.log(url.parse('amazon.com'));
console.log(url.parse('http://amazon.com'));
```

> Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: 'amazon.com',
  path: 'amazon.com',
  href: 'amazon.com'
}
Url {
  protocol: 'http:',
  slashes: true,
  auth: null,
  host: 'amazon.com',
  port: null,
  hostname: 'amazon.com',
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: 'http://amazon.com/'
}

As we observe, parsing without protocol results in different output.

I see the point of operation being done twice but resultant is different with(or out) protocol, If performance is the main concern we can parse the targetUrl before passing it to isMatchingUrl().

Would that be way to go?

sriv: wasn't thinking of performance, it's a simple operation when seen in context of the usecase.

I was thinking of the code duplication.

sriv:
@Manikkumar1988 - I've approved https://github.com/getgauge/taiko/pull/1299, but I'd like this to go in the release as well.

lib/handlers/targetHandler.js
Manikkumar1988: If Code duplication is of concern, not sure how do we fix that!

It looks hacky to me, may be we can do normal string comparison if `targetUrl` has prefix of 'http'? It is another workaround to surpass `url.parse()`

sriv: I get it. I think in that case we can live with this. String pattern match is worse.

sriv: I get it. I think in that case we can live with this. String pattern match is worse.

PR #1358, opened 2020-07-14
#279 Drag and drop support for html5 draggable elements

### LINKED ISSUE #279: Drag and drop does not work as expected for HTML5 draggable elements
### Expected behavior
Should be able to drag and drop HTML5 draggable elements.

### Actual behavior
Drag works fine but gets dropped to current cursor position instead of given position.

### Steps to reproduce
```
(async () => {
    try {
        await openBrowser();
        await goto("https://gojs.net/latest/samples/htmlDragDrop.html");
        await dragAndDrop("Coffee",into($('canvas')));
    } catch (e) {
        console.error(e);
    } finally {
        await closeBrowser();
    }
})();
```
Upstream issue for the same ChromeDevTools/devtools-protocol#130

NivedhaSenthil:
Fixes #279

### DIFF
--- a/lib/elements/element.js
+++ b/lib/elements/element.js
@@ -52,7 +52,7 @@ class Element {
       returnByValue: true,
     });
     if (result.value === undefined) {
-      throw new Error(`${this.description} does not have disable attribute`);
+      return false;
     }
     return result.value;
   }
--- a/lib/elements/element.js
+++ b/lib/elements/element.js
@@ -56,6 +56,20 @@ class Element {
     }
     return result.value;
   }
+
+  async isDraggable() {
+    function isDraggable() {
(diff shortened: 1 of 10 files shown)


lib/elements/element.js
```diff
@@ -52,7 +52,21 @@ class Element {
       returnByValue: true,
     });
     if (result.value === undefined) {
-      throw new Error(`${this.description} does not have disable attribute`);
+      return false;
+    }
+    return result.value;
+  }
+
+  async isDraggable() {
```
zabil, line 60: Can we document this API?

NivedhaSenthil: It has to be exposed from element wrapper also then, will do

Issue #1363, opened 2020-07-16
Text Improvement: English Language: Trivial Priority


davebneufeld:
The return value from the "exists" method would be better returning the text
"[PASS] Does not exist" versus the current text 
"[PASS] Does not exists" 

### Expected behavior

When testing if an element exists, the returning text should reflect that it is only one "thing"/element you are testing.
It would look better if it said: [PASS] Does not exist 

### Actual behavior
"[PASS] Does not exists"

### Steps to reproduce
1. add var NoSitesExistYet = text("[insert your text here]").exists(); to your script
2. run the script such that step 1 passes
3. profit!

### Versions
Version: 1.0.14 (Chromium: 85.0.4168.0)

NivedhaSenthil:
Thanks for the suggestion. Will be great to have a PR raised. Reference https://github.com/getgauge/taiko/blob/c63da878ec0f08bf0db08b1d163d2e118ae3ae20/lib/elementWrapper/elementWrapper.js#L27

NivedhaSenthil:
Fixed via #1372

PR #1372, opened 2020-07-20
Replace 'exists' return value to 'Does not exist'

felipdc:
#1363

Update 'exists' function in elementWrapper.js to return 'Does not exist'


NivedhaSenthil:
@felipdc Thanks for the contribution :+1:

Issue #1373, opened 2020-07-21
Memory leak on actions


NivedhaSenthil:
### Expected behavior
There should be no memory leaks happening for basic actions performed.

### Actual behavior
Memory leaks occurs after actions are peformed.

### Steps to reproduce
```
(async () => {
    try {
        await openBrowser();
        await goto("google.com");
        for(let i=0;i<1000;i++){
        await write("something");
        await clear();
        }
    } catch (error) {
        console.error(error);
    } finally {
        await closeBrowser();
    }
})();
```

Reference: https://spectrum.chat/taiko/general/running-gauge-tests-through-jenkins-failes~5ef7a885-7fba-4328-ab2a-dffe9664024b
### Versions
```
node - v14.4.0
taiko - 1.0.14
```

windyliu1118:
@NivedhaSenthil this fix didn't resolve the memory issue I faced, could you please take a look at my issue - https://spectrum.chat/taiko/general/taiko-node-consumes-too-much-memory-which-make-the-tests-running-terminated~f5384382-080e-4f85-8f2b-96c9524613ee

NivedhaSenthil:
@windyliu1118 can you please confirm if things are fine with the latest taiko 1.0.15

zabil:
Closing this as it's resolved for other users who have reported the issue.

ElmaCat:
I have encountered the heap memory issue: 
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
1: 00007FF70F695EBF napi_wrap+114095
2: 00007F
F70F640B46 v8::base::CPU::has_sse+66998
3: 00007FF70F641946 v8::base::CPU::has_sse+70582
4: 00007FF70FE56E4E v8::Isolate::ReportExternalAllocationLimitReached+94
5: 00007FF70FE3EF21 v8::SharedArrayBuffer::Externalize+833
6: 00007FF70FD0B18C v8::internal::Heap::EphemeronKeyWriteBarrierFromCode+1436
7: 00007FF70FD163C0 v8::internal::Heap::ProtectUnprotectedMemoryChunks+1312
Error ----------------------------------
Gauge version: 1.1.1
Commit Hash: 6e0d83f
Plugins
html-report (4.0.12)
js (2.3.12)
screenshot (0.0.1)
spectacle (0.1.4)
node -v
v12.18.3
taiko --version
[...]

PR #1374, opened 2020-07-21
#1373 clean up timeout

NivedhaSenthil:
- fix memory leak caused by timeouts


NivedhaSenthil:
Had done it on new call to the method, but it makes sense to do it on cleanup. Will push an update

PR #1407, opened 2020-08-04
Extend emulateNetwork to accept custom objects

carpeliam:
We have a few cases where we want to customize what throughput/latency values we want to use. With this PR, we can either use the preset values already present in Taiko, or supply an object that fits Chrome Dev Tools' needs. Hopefully this is still within the spirit of the repository, let us know if there are any changes you'd like us to make!

### DIFF
--- a/lib/handlers/networkHandler.js
+++ b/lib/handlers/networkHandler.js
@@ -59,7 +59,7 @@ const setNetworkEmulation = async (networkType) => {
   if (!networkType && _networkType) {
     networkType = _networkType;
   }
-  const emulate = networkPresets[networkType];
+  const emulate = (typeof networkType === 'string') ? networkPresets[networkType] : networkType;
   let networkModes = Object.keys(networkPresets);
   if (emulate === undefined) {
     throw new Error(`Please set one of the given network types \n${networkModes.join('\n')}`);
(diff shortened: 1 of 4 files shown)


zabil:
Thanks for the PR!

Is `{offline: true|false}` necessary? I am assuming the other values `{ downloadThroughput: number, uploadThroughput: number, latency: number }` are ignored when offline is true.

So the use case is see is

```
emulateNetwork('offline`)
```

or a custom

```
emulateNetwork({ downloadThroughput:  50, uploadThroughput: 2, latency: 2 }) // These are random numbers
```

carpeliam:
@zabil good question! We took a look at Chrome's [emulateNetworkConditions documentation](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-emulateNetworkConditions) to see if it handled default values at all, but it does not - I'm not sure how Chrome handles those if you omit a key. We like the idea though that something like `online` could be omitted and have a true value, so we've added default values of `{ offline: false, downloadThroughput: 0, uploadThroughput: 0, latency: 0 }` - so if I say `emulateNetwork({ downloadThroughput:  50, uploadThroughput: 2, latency: 2 })` or even `emulateNetwork({ downloadThroughput:  50, uploadThroughput: 2 })`, it would use these defaults and hopefully match user expectation. This includes an update to the documentation too.

PR #1410, opened 2020-08-05
Remove element wrapper list

### LINKED ISSUE #1341: elements(), element(index) and ElementWrapperList is missing from Typescript type definitions
# Expected behavior
`elements()` and `element(index)` from https://docs.taiko.dev/#elementwrapperlist is available in type definitions

### Actual behavior
`elements()`, `element(index)` and `ElementWrapperList` is not available in type definitions

### Steps to reproduce
Try to use `elements()` or `element(index)` in a typescript project.

### Versions
```
`taiko 1.0.12`
```

jfgreffier:
Fixes #1341

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -2973,7 +2973,7 @@ getPlugins().forEach((pluginName) => {
  * @param {string} optionName - Specifies the name of the configuration option/parameter you want to get (optional).
  * If not specified, returns a shallow copy of the full global configuration.
  * @param {string} string "navigationTimeout" Navigation timeout value in milliseconds for navigation after performing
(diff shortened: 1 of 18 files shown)


lib/taiko.js
```diff
@@ -3159,48 +3159,6 @@ const rectangle = async (selector, callback) => {
  *
  */
 
-/**
```
NivedhaSenthil, line 3162: These are properties of ElementWrapper, and the existing ElementWrapper definition has to be updated to Element. It has to be updated and not removed.

jfgreffier: Pushing corrections. Thanks for the pointer

I also fixed the types definitions the best I could by adding methods from different Elements, in the same manner that it was done with ElementWrapper

PR #1418, opened 2020-08-07
Print action output when env is set. #1328

### LINKED ISSUE #1328: Add an option to enable step description when using Jest as testrunner
### Expected behavior
It would be great, if "step description" could be enabled when using Jest as test-runner. (similar to the output from Taiko runner - see image).

### Actual behavior
When using Jest as testrunner, step description is currently not enabled (as opposed to Taiko runner which outputs information about every single step).

### Steps to reproduce
1. Clone https://github.com/getgauge-examples/jest-taiko
2. `npm i`
3. `npm run test`
4. Output contains only the testnames, not the step description

### Versions
```
node: v13.13.0
taiko: project local (latest)
```

BugDiver:
Fixes #1328

### DIFF
--- a/bin/taiko.js
+++ b/bin/taiko.js
@@ -102,6 +102,7 @@ function isCLICommand() {
 }
 
 if (isTaikoRunner(processArgv[1])) {
+  process.env.TAIKO_ENABLE_ACTION_OUTPUT = true;
   let plugins = getExecutablePlugins();
   if (
     isCLICommand() &&
--- a/bin/taiko.js
+++ b/bin/taiko.js
@@ -102,7 +102,7 @@ function isCLICommand() {
 }
 
 if (isTaikoRunner(processArgv[1])) {
-  process.env.TAIKO_ENABLE_ACTION_OUTPUT = true;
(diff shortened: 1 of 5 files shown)


bin/taiko.js
```diff
@@ -102,6 +102,7 @@ function isCLICommand() {
 }
 
 if (isTaikoRunner(processArgv[1])) {
+  process.env.TAIKO_ENABLE_ACTION_OUTPUT = true;
```
NivedhaSenthil, line 105: Has to be set only if not set by user

docs/configuring_taiko.md
```diff
@@ -2,13 +2,14 @@
...
 ---
 
-Listed below are instructions on how to modify Taiko's default paramaters
+Listed below are instructions on how to modify Taiko's default parameters
 for specific uses cases
 
 ## Using environment variables
 
-Taiko lets you specify certain Environment variables to customise its behaviour
+Taiko lets you specify certain Environment variables to customize its behavior
```
NivedhaSenthil, line 10: `behavior` has to be  `behaviour`

Issue #1062, opened 2020-01-29
How to use switchTo to switch between tabs with the same url and title?


filipesilva:
Hi there 👋 

### Expected behavior
I'd like to have a way to switch between tabs that can have the same title and url. This is useful for testing synchronization behaviour in a single-page application.

If I could choose a way to do it, I'd choose to use `switchTo(1)` where `1` is the tab number.

### Actual behavior
If I have two tabs with the same title and url, I cannot switch between them.

### Steps to reproduce
1. `npx taiko`
2. `openBrowser()`
3. `goto("https://www.google.com/")`
4. `openTab("https://www.google.com/")`
5. can't switch back to the first tab without closing the second, or changing its url/title

### Versions
```
taiko 1.0.3
node v12.4.0
```

BugDiver:
@filipesilva Switching tabs using index might not be the best way. Once the tab is closed it index has to updated  (reindex all the tabs)

Would opening a tab with an optional identifier, and then switch it using the same help your use case?

We are adding an `openWindow`API and we are adding the same options for that.

So I think we can do the same for tabs as well, and be consistent.

What do you think?

filipesilva:
Heya @BugDiver, thanks for getting back to me!

Yes I see the problem with indexes now that you mention it. An identifier sounds like the better approach.

Also happy to hear about `openWindow`, looking for that was what led me to using tabs instead!

BugDiver:
@filipesilva Thanks for confirming. I marking this as an enhancement. Also if you are willing to work on this we would be happy to help you.

## DEV NOTES

* Allow an option `identifier` to be passed to `openTab` API
* Use this id to mark the tab
* Allow an option `identifier` to be passed to `switchTo` API
* Use this to switch to the tab

filipesilva:
Heya, thanks for taking the time to implement a change for this feature!

PR #1432, opened 2020-08-18
Use identifiers for tabs

zabil:
Fixes: #1062
Fixes: ##1440

* target handler handles both window and tab identifier caching
* Logic for registering tabs and windows with windows is not out of `BrowserContext`

### Usage

```
openTab({ name: 'google'});
goto('google.com');
openTab('github.com', { name: 'github'});
openTab('taiko.dev')
switchTo({ name: 'github'});
switchTo(/google/);
closeTab({name: 'github'});
```

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -758,7 +758,8 @@ module.exports.closeTab = async (identifier) => {
   if (
     !identifier ||
     targetHandler.isMatchingUrl(activeTab, identifier) ||
-    targetHandler.isMatchingRegex(activeTab, identifier)
+    targetHandler.isMatchingRegex(activeTab, identifier) ||
+    targetHandler.isMatchingTarget(activeTab, identifier)
   ) {
     _client.removeAllListeners();
     await _client.close();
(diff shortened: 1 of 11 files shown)


lib/taiko.js
```diff
@@ -585,6 +587,13 @@ module.exports.openTab = async (
   if (!/^https?:\/\//i.test(targetUrl) && !/^file/i.test(targetUrl)) {
```
saikrishna321, line 587: @zabil Can we fail fast here.

```
  if (!options.name) {
    throw new TypeError('Tab name needs to be provided');
  }
```

saikrishna321 (changes requested):

lib/taiko.js
zabil: I did not add this check because unlike `incognitoWindow` a tab identifier using `options.name` is optional parameter. The user can choose to pass the identifier example calls would be

```
openTab('google.com')
openTab('GitHub.com', {name: 'GitHub'}):
```

Both should be allowed.

saikrishna321 (approved):

Issue #1383, opened 2020-07-28
tableCell missing from types


fmpanelli:
### Expected behavior
index.d.ts should export `tableCell` function

### Actual behavior
`tableCell` function is missing

### Steps to reproduce
1. open types/taiko/index.d.ts
2. search for tableCell

### Versions
```
node v12.16.1
taiko v1.0.16
```

NivedhaSenthil:
Thanks for reporting, will add them soon

fpanelli:
Thank you Nivedha for your prompt response!

fmpanelli:
Dear taiko team,

I had the opportunity to take a closer look at the topic.
I compared the functions declared in `index.d.ts` with the ones in `taiko.metadata` and I discovered that:

1. the following functions are exported by taiko.js, but they are missing from index.d.ts:
  - clearHighlights
  - color
  - getConfig
  - range
  - tableCell
  - timeField
2. A couple of functions are declared in index.d.ts, but are not exported from taiko.js:
  - intervalSecs
  - timeoutSecs

Additionally I tried to lint index.d.ts and got some warnings.

Now what... if you wish I can try to fix this, but it is somewhat invasive so I would like your ok before proceeding.

I am also trying to understand if it would be possible to introduce some testing or linting to avoid similar issues in the future (I am looking at dtslint and see if it can help).

What do you think?

Thanks
Filippo

NivedhaSenthil:
This is so great to see the effort you have put to gather these details, we would be very much happy to review a PR.

fmpanelli:
Ok 😀

I will start with a small PR to try the process and proceed incrementally.

Thank you,

fmpanelli:
This issue was resolved with PR #1454

PR #1435, opened 2020-08-19
Removed obsolete function defs

fmpanelli:
Ciao,

As discussed in issue #1383 there are a few inconsistencies in the "types" definition `index.d.ts`.

This PR simply removes from `index.d.ts` two functions (`intervalSec` and `timeoutSecs`) which were removed from `taiko.js` some months ago, but the corresponding type declaration has stayed untouched.

This is a very small PR just to get me started working on the topic.
Since this is my first PR I am new to this workflow, I am making this very small PR in order to check if everything is fine in the process. If everything's fine I'll propose some more beefy PR to address other types' inconsistencies and I will also try and introduce some testing for the types.

Thanks,
Filippo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1520,6 +1520,79 @@
         "to-fast-properties": "^2.0.0"
       }
     },
+    "@definitelytyped/header-parser": {
+      "version": "0.0.46",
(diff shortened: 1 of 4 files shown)


NivedhaSenthil (approved):

NivedhaSenthil:
@fmpanelli thanks for the contribution 👍

PR #1436, opened 2020-08-20
re-organize documentation, move it inline to types

sriv:


zabil:
Few observations.

* When the method does not return a value. It's better to say "This API does not return any values" over leaving it empty

Screenshot 2020-08-25 at 13 25 15


* The parameters get cramped up in lower resolution for example parameters for other methods
Screenshot 2020-08-25 at 13 33 13

and parameters for wrapper documentation

Screenshot 2020-08-25 at 13 33 41

Is there a way to make it consistent?

PR #1445, opened 2020-08-24
Fix text regression

### LINKED ISSUE #1442: Regression in text search / smart selectors
### Expected behavior
Functions using text search and smart selectors, like click(), should find specified text

### Actual behavior
click() using smart selectors does not find the text now, hence, can not click it

### Steps to reproduce
1) https://google-gruyere.appspot.com/start
2) Click "Agree & start"
3) Go to Sign up (/newaccount.gtl)
4) Create a random account
5) Run taiko:

```
> openBrowser()
[PASS] Browser opened
> goto("https://google-gruyere.appspot.com/.../login")
```

6) Log in with the above account
7) At this point, you should be at Gruyere Home, where you can see some links at the top. For example, "New Snippet"
7) In Taiko:

```
> click("New Snippet")
[FAIL] Error: Element with text New Snippet not found, run `.trace` for more info.
```

This works well in 1.0.18 but 1.0.19 is breaking some of my tests. May be related to https://github.com/getgauge/taiko/pull/1429

v14.7.0
> taiko --version
Version: 1.0.19 (Chromium: 86.0.4214.0) RELEASE

NivedhaSenthil:
fixes #1442

### DIFF
--- a/lib/elementSearch.js
+++ b/lib/elementSearch.js
@@ -10,6 +10,7 @@ function match(text, options = {}, ...args) {
   const get = async (tagName = '*') => {
     let elements;
     const textSearch = function (selectorElement, args) {
+      const searchText = args.text.toLowerCase().trim();
       const nodeFilter =
         args.tagName === '*'
           ? {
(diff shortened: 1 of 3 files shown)


lib/elementSearch.js
```diff
@@ -10,6 +10,7 @@ function match(text, options = {}, ...args) {
   const get = async (tagName = '*') => {
     let elements;
     const textSearch = function (selectorElement, args) {
+      const searchText = args.text.toLowerCase().trim();
```
zabil, line 13: Do we need a null check here?

NivedhaSenthil: we do a type assertion at the beginning of method.. https://github.com/getgauge/taiko/blob/2b9e3dc11c887b0bfb5560a043600e742c3617c2/lib/elementSearch.js#L9

zabil (approved):

PR #1462, opened 2020-08-28
Test types for browser actions

fmpanelli:
Hi!

With this PR I submit tests for all types related to taiko API's browser actions, i.e.:

- openBrowser
- closeBrowser
- client
- switchTo
- intercept
- emulateNetwork
- emulateDevice
- setViewPort
- openTab
- closeTab
- openIncognitoWindow
- closeIncognitoWindow
- overridePermissions
- clearPermissionOverrides
- setCookie
- deleteCookies
- getCookies
- setLocation
- clearIntercept

Some of them required to fix the type declaration in `index.d.ts`.

Also the last commit imports (as dev dependency) `devtools-protocol` which is the official repository of the protocol and provides some useful types that otherwise we should re-define. In this commit we use the type `Cookie` for `getCookies`. previously that was treated simply as an `object`, now it gets proper typing.

Sorry if the PR is quite _heavy_ and it may require some time to review.

Filippo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4754,6 +4754,12 @@
       "integrity": "sha1-p2o+0YVb56ASu4rBbLgPPADcKPA=",
       "dev": true
     },
+    "devtools-protocol": {
+      "version": "0.0.802093",
+      "resolved": "https://registry.npmjs.org/devtools-protocol/-/devtools-protocol-0.0.802093.tgz",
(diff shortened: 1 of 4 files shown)


NivedhaSenthil (approved):

NivedhaSenthil:
@fmpanelli Thanks for taking effort to correct these 👍

PR #1496, opened 2020-09-10
add eslint for no-instanceof and fix some

NivedhaSenthil:


lib/elementSearch.js
```diff
@@ -180,7 +180,7 @@ const $$xpath = async (selector, selectHiddenElements) => {
 
 const $function = async (callBack, args, selectHiddenElements) => {
   function searchThroughShadowDom(argument) {
-    const isString = (obj) => typeof obj === 'string' || obj instanceof String;
+    const isString = (obj) => Object.prototype.toString.call(obj).includes('String');
```
zabil, line 183: Can't we use or assign the `isString` in the `helper.js`?

NivedhaSenthil: this function is run inside browser context, hence its an overhead passing another simple method like string check to it

PR #1500, opened 2020-09-11
#1476 fix actions on shadowdom

### LINKED ISSUE #1476: button click is clicking on an incorrect element for version 21,20,19 works for v18
### Expected behavior
Button Click with Text displayed should click on the right element. 
UI Element - 
image

DOM -
image

### Actual behavior
1. Using 
await highlight(button("Submit"));
await click(button("Submit"));

clicks on the following element
UI -
image

DOM-
image

I have tried with some of the taiko versions
1. v1.0.21 - Fails
2. v1.0.20 - Fails
3. v1.0.19 - Fails
4. v1.0.18 - Works

But this works in REPL for all versions, but when using it with Gauge scripts in a step definitions, it fails and clicks on the incorrect element. 

### Versions
node - v12.16.1
Gauge version: v1.1.1

NivedhaSenthil:
- there can be cases where element at point being outside the shadowdom in which case recursive piercing can go infinity
- bump up version to 1.0.23 for release
- hence pierce through each shadow root only once

Fixes #1476

### DIFF
--- a/lib/actions/dragAndDrop.js
+++ b/lib/actions/dragAndDrop.js
@@ -56,14 +56,18 @@ const dragAndDropHTML5Element = async (sourcePosition, destPosition, sourceEleme
   function simulateDragAndDrop(args) {
     let sourceElement, targetElement;
     sourceElement = document.elementFromPoint(args.sourcePosition.x, args.sourcePosition.y);
-    while (sourceElement.shadowRoot) {
+    let parsedShadowRoots = [];
+    while (sourceElement.shadowRoot && !parsedShadowRoots.includes(sourceElement.shadowRoot)) {
+      parsedShadowRoots.push(sourceElement.shadowRoot);
       sourceElement = sourceElement.shadowRoot.elementFromPoint(
         args.sourcePosition.x,
         args.sourcePosition.y,
       );
     }
     targetElement = document.elementFromPoint(args.destPosition.x, args.destPosition.y);
-    while (targetElement.shadowRoot) {
+    parsedShadowRoots = [];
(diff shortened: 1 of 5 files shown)


sriv:
This pattern is repeated:

```
    let parsedShadowRoots = [];
    while (sourceElement.shadowRoot && !parsedShadowRoots.includes(sourceElement.shadowRoot)) {
      parsedShadowRoots.push(sourceElement.shadowRoot);
      ...
```

Can we have this extracted to a function ?

lib/actions/dragAndDrop.js
```diff
@@ -56,14 +56,18 @@ const dragAndDropHTML5Element = async (sourcePosition, destPosition, sourceEleme
   function simulateDragAndDrop(args) {
     let sourceElement, targetElement;
     sourceElement = document.elementFromPoint(args.sourcePosition.x, args.sourcePosition.y);
-    while (sourceElement.shadowRoot) {
+    let parsedShadowRoots = [];
```
sriv, line 59: This pattern is repeated:

```
    let parsedShadowRoots = [];
    while (sourceElement.shadowRoot && !parsedShadowRoots.includes(sourceElement.shadowRoot)) {
      parsedShadowRoots.push(sourceElement.shadowRoot);
      ...
```

Can we have this extracted to a function ?

sriv (changes requested):

lib/actions/dragAndDrop.js
NivedhaSenthil: since these functions are executed separately in browser context we will have to pass the extracted common function as args, and those arg will be serialised as string so to again use it we will have to convert them back to function.. I feel it will make it more complex making it less readable

sriv (approved):

PR #1535, opened 2020-09-23
Add regex support for text() API

### LINKED ISSUE #286: Regex match of text
**Expected behavior**
Should be able to if a text that matches a regex is present.
For example:- `Some text <timestamp> Some other text`

**Actual behavior**
Currently regex option cannot be specified to the text

**Proposed behavior**
The text API can be overloaded with another method which takes regex as a parameter
`text(/regex/)` for matching the text.

**Version**
```
commit - b7f2079445fe01fbd6cfe08babf1ed7ba52be6a0
```

NivedhaSenthil:
Fixes #286


lib/elementSearch.js
```diff
@@ -31,52 +41,65 @@ function match(text, options = {}, ...args) {
...
       const iterator = document.createNodeIterator(
         selectorElement,
         NodeFilter.SHOW_ALL,
         nodeFilter,
       );
+
       const exactMatches = [],
         containsMatches = [];
 
+      function checkIfRegexMatch(text, exactMatch) {
```
zabil, line 54: I am finding it hard to understand this method especially because this works on a variable `searchText` which is outside it's scope. I recommend passing it as an input parameter.

zabil (changes requested):

types/taiko/test/selectors.ts
```diff
@@ -173,3 +173,7 @@ text('Vehicle'); // $ExpectType TextWrapper
 text('Vehicle', below('text')); // $ExpectType TextWrapper
 text('Vehicle', { exactMatch: true }, below('text')); // $ExpectType TextWrapper
 text('Vehicle', { selectHiddenElements: true }); // $ExpectType TextWrapper
+text('Vehicle', { selectHiddenElements: true }); // $ExpectType TextWrapper
+text('/Vehicle/'); // $ExpectType TextWrapper
```
zabil, line 177: I don't think regex as string should be supported. This can be a valid text which will then be considered a regex leading to unwanted matches and escaping. I think it is ok to explicitly use `/test/` or `new Regexp` to indicated a search by regex.

NivedhaSenthil: It is actually the side effect of serializing object when running on browser to find  regex matches, even the Regex object is converted to string. Just wanted to explicitly call out so that users can know how a string starting and ending with a `/` is treated

zabil (approved):

Issue #1530, opened 2020-09-21
API reference lacks return type for some functions


fmpanelli:
### Expected behavior
Pages in [API reference](https://docs.taiko.dev/api/reference) should always show the return type if it is included in the JSDoc comment of the API. 

E.g. `goto`  in the JSDoc has:
```javascript
 * @returns {Promise<Object>} response
```

therefore the [refernce page](https://docs.taiko.dev/api/goto/) should show something like: 

```
Return
Promise<Object>
response
```

### Actual behavior

Instead the [refernce page for goto](https://docs.taiko.dev/api/goto/) shows:

![image](https://user-images.githubusercontent.com/53738017/93735948-f8944700-fbde-11ea-9202-5826611b1464.png)

### Steps to reproduce
1. open chrome
2. type url `https://docs.taiko.dev/api/goto/`
3. compare with JSDoc comment in `taiko.js`

### Versions
```
node v14.4.0
taiko 1.0.23
vscode 1.49.0
```

fmpanelli:
Closing after PR #1538 was merged.

Now the page at https://docs.taiko.dev/api/goto/ says:
![image](https://user-images.githubusercontent.com/53738017/94957519-2dad6d00-04ee-11eb-9d29-a83a61e93cb5.png)

PR #1538, opened 2020-09-24
Fixing Issue 1530

fmpanelli:
This PR contains quite a few things:

- documenting/explaining the process for generating documentation
- setting up a harness to test the documentation-generation process and `njk` files in particular
- fixing issue #1530 with `Promise<Object>` return type

Thank you in advance for reviewing it.

Filippo


docs/README.md
```diff
@@ -9,3 +9,55 @@ npm run doc
...
+2. it runs [eleventy](https://www.11ty.dev/) to generate the website
+
+#### Step 1 - Generating the JSON
+
+This steps uses the [documentation.js](https://documentation.js.org/)  library to parse the JSDoc comments in the code and generate a JSON file name `lib/api.json`.
+
+The chain is the follwing:
+
+- you digit `npm run doc`
+- as a first step this runs `npm run doc:api`
```
zabil, line 28:
```suggestion
- As a first step this runs `npm run doc:api`
```

docs/README.md
```diff
@@ -9,3 +9,55 @@ npm run doc
...
+- this launches `lib/documentation.js`
```
zabil, line 29:
```suggestion
- Which launches `lib/documentation.js`
```

docs/README.md
```diff
@@ -9,3 +9,55 @@ npm run doc
...
+npx documentation build --shallow lib/taiko.js lib/elementWrapper/*.js -o lib/api.json
+```
+
+As you can see, at the timw of writing, only some of the code is parsed:
+- `taiko.js` contains the definition of the APIs for external use
+- `lib/elementWrapper/*.js` contains the definition of some classes (`ElementWrapper` and company) which are used as return type for api's functions
+
+#### Step 2 - Generating the web site
+
+The second step uses [eleventy](https://www.11ty.dev/)  to generate the web site. There is a `.eleventy.js` file in the project's root dir which configures the tool and in particular sets:
```
zabil, line 43:
```suggestion
The second step uses [eleventy](https://www.11ty.dev/) to generate the web site. There is a `.eleventy.js` file in the project's root dir which configures the tool and in particular sets:
```

docs/api/api.njk
```diff
@@ -1,12 +1,29 @@
...
-permalink: "api/{{ api.name }}/"
-layout: api.njk
+    data: apis.functions 
+    size: 1 
+    alias: api 
+permalink: "api/{{ api.name }}/" 
+layout: api.njk 
 ---
 
+{% macro simpleTypeParser(type) -%}
```
zabil, line 10: can we call this macro 
```suggestion
{% macro linkToType(type) -%}
```
?

lib/jsDocToJson.js
```diff
@@ -0,0 +1,21 @@
+const exec = require('child_process').exec;
```
zabil, line 1: Is there a specific reason why this has been extracted to a separate file? Can we instead have `jsDocTJson` inside `documentation.js` that way it would be easy to use it later for example.

```
let doc = require('documentation');
doc.jsDocToJson(...);
```
What do you think?

test/docs-tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+1. generate a test website
+2. test the website using taiko
+
+Currently the test is focused on the Nunjucks `njk` templates.
+
+### Generating the test website
+
+The test command, as the first step, generates a test website by:
+
+- copying the structure from the `docs` dir in the project root and in particular the `njk` files
```
zabil, line 26:
```suggestion
- Copying the structure from the `docs` dir in the project root and in particular the `njk` files
```

test/docs-tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+- running `documentation.js` on the test files in the `inputs` folder to generate an `api.json` file
```
zabil, line 27:
```suggestion
- Running `documentation.js` on the test files in the `inputs` folder to generate an `api.json` file
```

test/docs-tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+- running `eleventy.js` on the generated `api.json` file with the copied `njk` files
```
zabil, line 28:
```suggestion
- Running `eleventy.js` on the generated `api.json` file with the copied `njk` files
```

zabil (changes requested):
Thanks for this pull request!

Added a few comments on typo and a few suggestions. Tell me what you think about it.

lib/jsDocToJson.js
fmpanelli: I already tried as you suggest, but I encountered a problem that made me split the files.
Maybe you have a better way to address it.

 The problem is that `documentation.js` contains also the following code at the top level:
```javascript
if (!process.env.TAIKO_SKIP_DOCUMENTATION) {
  const sourceFiles = 'lib/taiko.js lib/elementWrapper/*.js';
  const outputFile = 'lib/api.json';
  jsDocToJson(sourceFiles, outputFile);
}
```

So when I do `let doc = require('documentation');` in the file `test/docs-tests/generateWebsite.js`, the code above is immediately run.
The net result is that `jsDocToJson` is executed twice 😮

Splitting the files allowed me to avoid importing the top-level code of `documentation.js`. Do you have in mind any other option to fix that?

zabil: Ah ok, now I see. You can try this

```
if (require.main === module && !process.env.TAIKO_SKIP_DOCUMENTATION) {
   const sourceFiles = 'lib/taiko.js lib/elementWrapper/*.js';
   const outputFile = 'lib/api.json';
   jsDocToJson(sourceFiles, outputFile);
 }
```

docs/api/api.njk
fmpanelli: Yes, I will refactor the code and submit a new commit.

lib/jsDocToJson.js
fmpanelli: Cool!

I did not know. I keep learning a lot, thank you! 😀

fmpanelli:
Thank you Zabil for the review and for the many typos+fixes.
I think I have incorporated all your suggestions.

(I messed up a little with signing the commits, but it should be fine now)

package.json
```diff
@@ -20,6 +20,7 @@
     "test:unit": "mocha 'test/unit-tests/**/*.test.js' --timeout 6000 --trace-warnings",
     "test": "npm run test:api && npm run test:unit:silent",
     "test-functional": "cd test/functional-tests && npm install && npm test",
+    "test-docs": "cd test/docs-tests && node generateWebsite.js && cd ./gauge && npm install && npm test",
```
zabil, line 23: Before merging I wanted to check where the test is run. I think we should run it after a check-in via GitHub actions. However I do not see a workflow entry for running it.

fmpanelli: Yes.
Currently you can run it manually like `npm run test-docs`, with no automation.

If I understand correctly, you are suggesting to insert in `.github/workflows/taiko` an additional job that launches this test (which I have never done, but I can try). Is my understanding correct?

Should we instead or in addition add an `husky` hook before committing or pushing?
Like 
 ```json
//package.json
...
 "husky": {
    "hooks": {
      "pre-commit": "lint-staged && npm run test:api && npm run test-docs",
      "pre-push": "npm test"
    }
  },
...
```

zabil: Just adding it to `.github/workflows/taiko` is fine for now as these are functional tests. The husky pre-commit hook currently only runs unit tests.

fmpanelli: I tried several times, but it looks like I am not able... maybe I'll try again tomorrow.

fmpanelli: My impression is that CI/CD pipeline is broken for all PRs, even when I rolled back to `e12a2e0` which previously was ok...

image

I am not able to reproduce the same problem on my machine. So I don't know what to do about that,,,

fmpanelli: ok, I was able to reproduce it locally... opening an issue right away

zabil: A fix was pushed for this, can you try rebasing and adding an entry to `.github/workflows/taiko`?

fmpanelli: I rebased to clean up the history which was getting somewhat contorted.

Then I applied to `docs-tests` similar changes to those made by BugDiver in `functional-tests`.

Finally, I am trying to add the new job to the github workflow 🤞

fmpanelli: All green 🎉

zabil (approved):
👏🏻

Issue #1499, opened 2020-09-10
The Taiko selectors checkBox and radioButton do not bubble the click event.


falcon030:
### Expected behavior
When I use radioButton().select() or checkBox.check() I expect the JavaScript `click` event to be triggered.

### Actual behavior
The JavaScript click event is not triggered, when  using radioButton().select() or checkBox.check().

This is probably caused by the fact that the setChecked function of the `radioButton` and `checkBox` API does not bubble the `click` event:
```
function setChecked(value) {
  this.checked = value;
  let event = new Event('click');
  this.dispatchEvent(event);
}

Taiko version 1.0.19
```

saikrishna321:
Fix in master.

PR #1542, opened 2020-09-25
#1499 bubble events for checkBox and radioButton API

saikrishna321:
Addresss #1499


NivedhaSenthil (changes requested):
Can some unit tests be added for this please ?

Issue #1552, opened 2020-09-29
Functional test failing with Gauge ts plugin ver 0.1.0


fmpanelli:
Running `npm run test-functional` crashes if you do not have the correct version of gauge ts plugin  (you can see all GitHub actions failed in the last few days).

### Expected behavior
`npm run test-functional` should work and print out results of the functional tests.

### Actual behavior
If you do not have a `ts` plugin already installed, when you execute `npm run test-functional` it installs the version `0.1.0` and then crashes with a horrible message 😱😱

![image](https://user-images.githubusercontent.com/53738017/94591296-2f75f700-0288-11eb-9fbc-e0f647985ee3.png)

### Steps to reproduce
1. delete the content of your gauge plugin folder (in my case `rm -rf ~/.gauge/plugins/`)
2. go to your `taiko` folder
3. execute  `npm run test-functional` 

To restore the correct behaviour 
1. delete the content of your gauge plugin folder (in my case `rm -rf ~/.gauge/plugins/`)
2. execute `gauge install ts -v 0.0.5`
3. go to your `taiko` folder
4. execute  `npm run test-functional` 

### Versions
```
node v14.8.0
taiko v 1.0.23
```

zabil:
Looks like an issue with the newer changes to the Typescript plugin https://github.com/BugDiver/gauge-ts/releases/tag/v0.1.0

fmpanelli:
See PR #1553

As a temporary fix I've forced functional-tests to install the version 0.0.5 of the plugin and uninstall version 0.1.0 if present.

At least now the pipeline is green again!

Then somebody will have to understand why the newer version is crashing the tests.

fmpanelli:
Closing after BugDiver fixed the root cause with commit 406b944

PR #1553, opened 2020-09-29
Forcing gauge ts plugin version to 0.0.5

fmpanelli:
This is a temporary fix, because the new version of the plugin 0.1.0
crashes the functional tests and the GitHub pipeline.

it should fix issue #1552


BugDiver:
@fmpanelli This PR would be required anymore. The FT's has been fixed and working properly with the latest typescript plugin. Closing it.

PR #1563, opened 2020-10-01
Add support to get all values from drop down

saikrishna321:
New API to return an array with all the options from a dropdown

```
await dropDown('Vehicle:').options()
```


NivedhaSenthil:
I feel this can be named `options` instead of `values` as value is generally linked to selected option of the dropdown.

zabil:
> I feel this can be named `options` instead of `values` as value is generally linked to selected option of the dropdown.

Ah yes, I agree. `options` is the right name for this API and how users refer to it.

zabil:
>  I feel this can be named `options` instead of `values` as value is generally linked to selected option of the dropdown.

@saikrishna321 ran into an issue with the name `options` this is because 

https://github.com/getgauge/taiko/blob/c082b8c4060cbab4484d3741fae9c55da95d0725/lib/elementWrapper/elementWrapper.js#L27-L30

sets the `options` variable so calling `dropDown({ id: 'select' }).options()` will refer to `this.options` variable instead. We should either change `const { selector, options } = prepareParameters(attrValuePairs, _options, ...args); ` to something else or call the method `dropDown({ id: 'select' }).valueOptions()`

lib/taiko.js
```diff
@@ -2488,6 +2488,8 @@ function textBox(labelOrAttrValuePairs, _options, ...args) {
  * @example
  * await dropDown('Vehicle:').value()
  * @example
+ * await dropDown('Vehicle:').values() - Returns all the values from the drop down
```
zabil, line 2491: shouldn't this be 
```suggestion
 * await dropDown('Vehicle:').options() - Returns all available options from the drop down
```

lib/elements/dropDown.js
```diff
@@ -96,6 +96,21 @@ class DropDown extends Element {
     });
     return result.value;
   }
+
+  async options() {
+    function getValues() {
```
zabil, line 101:
```suggestion
    function getOptions() {
```

lib/elements/dropDown.js
```diff
@@ -96,6 +96,21 @@ class DropDown extends Element {
...
+      let dropDownValues = [];
+      for (let index = 0; index < this.options.length; index++) {
+        dropDownValues.push(this.options[index].value);
+      }
+      return dropDownValues;
+    }
+    const { result } = await this.runtimeHandler.runtimeCallFunctionOn(getValues, null, {
```
zabil, line 108:
```suggestion
    const { result } = await this.runtimeHandler.runtimeCallFunctionOn(getOptions, null, {
```

test/unit-tests/dropDown.test.js
```diff
@@ -132,6 +132,15 @@ describe(test_name, () => {
       expect(await dropDown('Cars').value()).to.equal('mercedes');
     });
 
+    it('test values()', async () => {
```
zabil, line 135:
```suggestion
    it('test options()', async () => {
```

zabil (changes requested):

zabil (approved):

PR #1602, opened 2020-10-16
Add support for multi-select Drop downs

### LINKED ISSUE #1543: Multiple Value select for dropDown API
This is a feature request for selecting multiple values with the dropDown API

### Expected behavior
dropDown(label).select(option1,option2') ==> Should select multiple values
dropDown(label).select({index:0,1,2,3})
dropDown(label).select(/subString/) ==> Should select all the options which match the given Regex

### Actual behavior
dropDown(label).select(option1,option2') ==> throws error
dropDown(label).select({index:0,1,2,3}) ==> this selects only the last option ie index 3
dropDown(label).select(/subString/) ==> selects first matching regex

luthrap:
Fixes #1543 |Paras+Jyoti|


lib/elementWrapper/dropDownWrapper.js
```diff
@@ -51,12 +51,12 @@ class DropDownWrapper extends ValueWrapper {
   }
 
   /**
-   * Selects the option corresponding to value passed.
-   * @param {any} value value to be selected
+   * Selects the option corresponding to value passed. Select multiple options if array of values passed
+   * @param {any} values value(s) to be selected
```
zabil, line 55:
```suggestion
   * @param {string[]} values value(s) to be selected
```

lib/elementWrapper/dropDownWrapper.js
```diff
@@ -51,12 +51,12 @@ class DropDownWrapper extends ValueWrapper {
   }
 
   /**
-   * Selects the option corresponding to value passed.
-   * @param {any} value value to be selected
+   * Selects the option corresponding to value passed. Select multiple options if array of values passed
```
zabil, line 54:
```suggestion
   * Select option or options from a drop down
```

zabil (changes requested):
This is not a full review.

Notice that the docs can be modified to reflect the right parameter type.

luthrap:
Will do that. Also, had to add some extra conditions to support the test case in unit-tests/elements/dropDown.test.js. The conditions were added as we have changed the CDP command to select a drop down value to support multi-select which was not getting supported in the current node structure.

lib/elements/dropDown.js
```diff
@@ -30,66 +34,164 @@ class DropDown extends Element {
...
     const options = setNavigationOptions({});
 
     // RegExp is not serialised. So loop through dropdown text and find a match
     // before invoking via CDP API
-    let valueToSelect = value;
-    if (isRegex(value)) {
+
+    let valuesToSelect = values;
+    if (Array.isArray(values)) {
+      //do nothing
```
saikrishna321, line 111: Dead code. Can this be removed?

lib/elements/dropDown.js
```diff
@@ -30,66 +34,164 @@ class DropDown extends Element {
...
+    } else if (typeof values == 'undefined') {
+      valuesToSelect = [undefined];
+    } else if (!values) {
+      valuesToSelect = [null];
```
saikrishna321, line 115: What is the need to set null and undefined?

lib/elements/dropDown.js
```diff
@@ -30,66 +34,164 @@ class DropDown extends Element {
...
+        if (this.options[index].selected || this.selectedIndex === index) {
+          if (this.options[index].value) {
+            selectedValues.push(this.options[index].value);
+          } else {
+            selectedValues.push(this.options[index].innerText);
+          }
+        }
       }
-      return this.innerText;
+      return selectedValues.join(',');
```
saikrishna321, line 194: Why can't we just return the array `selectedValues `?

saikrishna321 (changes requested):
Few changes needed!

lib/elements/dropDown.js
luthrap: Although it was written to skip other elseif conditions if the value provided is an array, will refactor it.

lib/elements/dropDown.js
luthrap: If we just return array, it will not be backward compatible for which single value was provided.

saikrishna321: Thought - Should we have a new function called `values()` which would return multiple values in array format. 
@NivedhaSenthil @zabil

lib/elementWrapper/dropDownWrapper.js
```diff
@@ -51,12 +51,12 @@ class DropDownWrapper extends ValueWrapper {
   }
 
   /**
-   * Selects the option corresponding to value passed.
-   * @param {any} value value to be selected
+   * Select option or options from a drop down
+   * @param {string[]} values value(s) to be selected
```
zabil, line 55: Can you also add a few examples usages of using array as parameters? You can add it here using the `@example` tag and at `dropDown` API jsdoc which will then appear at https://docs.taiko.dev/api/dropdown/

lib/elements/dropDown.js
luthrap: Fixed. Please check.

lib/elementWrapper/dropDownWrapper.js
luthrap: Added.

lib/elements/dropDown.js
luthrap: Fixed.

lib/elements/dropDown.js
zabil: > Thought - Should we have a new function called values()

You mean selected values right? Yeah, that makes sense. Can be part of another PR. @saikrishna321 can you create an issue for that?

saikrishna321: @zabil Yes, for selected values. Will create a new issue for that.

saikrishna321: @luthrap We can address this here  https://github.com/getgauge/taiko/issues/1613. Can this be reverted?

luthrap: Sure, will revert and update the test cases. Meanwhile, it appears that a lot of extra files have been checked in as part of the PR. This happened while signing off the previous commits, I mistakenly signed in those commits that should not belong to this PR. Could it be removed from here or shall i open a new clean PR request.

zabil: > Sure, will revert and update the test cases.

Sorry, if it's been already implemented let's keep it! Maybe refer to the issue in this PR

> Could it be removed from here or shall i open a new clean PR request.

I noticed that too. You could update the same PR by fixing the merge and arranging the commits.

saikrishna321: >if it's been already implemented let's keep it

@zabil It's done as a part of current function `value()`

luthrap: We can move the value() code of the PR request in the values() function and have the original logic in the value() function. This will require a change in a couple of test cases and would conclude the PR correctly.

zabil: > It's done as a part of current function value()

IMHO that should be fine. It might be confusing to have two different API's to return values depending on the type of the box when there is a common API to set the values depending on the type of the combo box.

So I am assuming `value` returns and array when there are multiple values selected

@saikrishna321 luthrap What do you think?

luthrap: Agreed, we should be encapsulating the logic in one function only.
```
if(selectedValues.length <= 1){
  return selectedValues.join()
else{
  return selectedValues
}
```

saikrishna321: > So I am assuming value returns an array when there are multiple values selected

Agree!

lib/elements/dropDown.js
```diff
@@ -30,66 +34,160 @@ class DropDown extends Element {
...
+        if (!result.value[key].isAvailable) {
+          unavailableValues.push(key);
+        }
+      });
+
+      if (disabledValues.length > 0) {
+        throw new Error(`Cannot set value ${disabledValues.join(',')} on a disabled field`);
+      } else if (Object.keys(result.value).length === 0) {
+        if (typeof values === 'undefined') {
+          throw new Error('Option undefined not available in drop down');
```
saikrishna321, line 152: Can you add a test to validate this?

saikrishna321 (changes requested):

lib/elements/dropDown.js
luthrap: The tests has already been already in dropDown.test.js. For the disabled field, following test are added :
```
 it('test select() should throw when selecting multiple values and one of the value is disabled', async () => {
      try {
        expect(await dropDown('Foods').select(['Burger', 'Pasta']));
      } catch (err) {
        expect(err.message).to.equal('Cannot set value Burger on a disabled field');
        expect(await dropDown('Foods').value()).to.equal('');
      }
    });
```

Test for undefined was already present earlier in the master branch 
```
it('should throw error for undefined values', async () => {
      try {
        expect(await dropDown('Cars').select(undefined));
      } catch (err) {
        expect(err.message).to.equal('Option undefined not available in drop down');
      }
    });
```

zabil (approved):

zabil:
@saikrishna321 will wait for your approval.

The PR looks good to me. Great work! @luthrap @guptajyoti845

luthrap:
@zabil : I am yet to commit the changes for `value()` function. Will do it by EOD. If @saikrishna321 has a go ahead, PR can be merged once i update the `value()` function.

saikrishna321:
@luthrap @guptajyoti845 Thanks a lot for working on all the reviews. 

@zabil LGTM! @luthrap Once the `value()` is fixed. Happy to merge

luthrap:
Pushed the changes.

**Update** 
If we look at the signature below, it returns a `Promise<string>`. However, we should return `Promise<string | Array<string> | Array<number>>` as this is the correct way. Should we add this? Although adding this will start breaking other tests from `htmlElementAPI.ts`
```
export interface ValueWrapper extends ElementWrapper {
  value(): Promise<string>;
}
```

zabil:
> However, we should return `Promise<string | Array<string> | Array<number>>` as this is the correct way although it does not impact the functionality. Should we add this? Adding this will start breaking other tests from `htmlElementAPI.ts`.

Adding @fmpanelli to the discussion.

saikrishna321 (approved):

luthrap:
@zabil @saikrishna321 : can it be merged and the issue related to the change in contract can be picked up in other issue.

zabil:
Agree. I'll test this a bit more today and merge it.

fmpanelli:
I am having a look at the types...

PR #1635, opened 2020-10-28
Update doc to use index as parameter to elements

zabil:


docs/working_with_element_lists.md
```diff
@@ -38,27 +38,32 @@ await listItem({class: "fruit"}).text();
...
-returns an array of matches
+To access other matches of the list items with the class attribute `fruit` 
+you can use the [`elements`](https://docs.taiko.dev/api/elementwrapper/#elements) 
+in two ways.
+
+By passing the index of the match you want to fetch
 
 ```
-var fruits = await listItem({class: "fruit"}).elements(); 
+var fruit = await listItem({class: "fruit"}).elements(0);
```
saikrishna321, line 53: Should be: `const fruit = await listItem({class: "fruit"}).element(0)`

saikrishna321 (changes requested):

docs/working_with_element_lists.md
zabil: Thanks! pushed the fix.

saikrishna321 (approved):

PR #1636, opened 2020-10-28
cookie consent breaks the script when not run headless

gnoireaux:
For people trying out Taiko, it might be useful to have a script that doesn't break when they observe it.

### DIFF
--- a/docs/record_and_run_tests.md
+++ b/docs/record_and_run_tests.md
@@ -89,6 +89,7 @@ Google
 
 ```
 > goto("google.com")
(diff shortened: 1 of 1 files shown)


zabil:
The PR looks good!

Can you sign the commit so that the DCO check passes? You can find more details about that under "signing commits" at https://docs.taiko.dev/contributing/#contribute-code-and-pull-requests-to-taiko!

gnoireaux:
DCO passes on https://github.com/getgauge/taiko/pull/1637

PR #1637, opened 2020-10-28
cookie consent breaks the script when not run headless

gnoireaux:
For people trying out Taiko, it might be useful to have instructions and a script that do not break when they observe it.

Signed-off-by: Guillaume Noireaux <gnoireaux@gmail.com>


gnoireaux:
Thanks Zabil

PR #1641, opened 2020-10-30
Handling empty body in mock response #1596

### LINKED ISSUE #1596: Empty body for intercept not possible
**Describe the bug**
If not setting the body for an intercept, the intercept fails when called.
There are responses without bodies, e.g. 204 No Content, so this behaviour has to be an expected one.
Same applies if the body actively is set to `undefined` or `null`

**To Reproduce**
```
const { openBrowser, intercept, goto } = require('taiko');
(async () => {
  await openBrowser();
  await intercept('https://example.com', {status: 204});
  await goto('https://example.com');
})()
```

**Error message**

```
Uncaught The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined
```

**Expected behavior**
Intercept works.

**Additional context**
* This bug is caused by `Buffer.from(responseBodyJson)` in `lib/handlers/fetchHandler.js` (undefined is passed to Buffer.from)

BugDiver:
Fixes #1596

### DIFF
--- a/lib/handlers/fetchHandler.js
+++ b/lib/handlers/fetchHandler.js
@@ -119,7 +119,7 @@ const handleInterceptor = (p) => {
 
 const mockResponse = (response, options) => {
   let responseBodyJson = isObject(response.body) ? JSON.stringify(response.body) : response.body;
-  const responseBody = Buffer.from(responseBodyJson);
+  const responseBody = Buffer.from(responseBodyJson || '');
 
   const responseHeaders = {};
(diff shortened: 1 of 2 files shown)


zabil:
Can we add a test for this?

BugDiver:
> Can we add a test for this?

Done

saikrishna321 (approved):

NivedhaSenthil (approved):

PR #1648, opened 2020-10-31
Change lib/taiko adds omitDescription option in evaluate

### LINKED ISSUE #806: Provide a option to omit descriptions in repl
### Expected behavior
There has to be a way to omit the descriptions in taiko APIs

### Actual behavior 
Currently, the user can not omit the descriptions given from the APIs

### Use Case
In REPL mode all the taiko APIs prints a description about the action they performed. The plugins can also contribute to this feature by using the description event emitter which is passed to the plugins.
But when the plugin is using any of the taiko APIs within its own  APIs, The description of the taiko APIs also gets printed and that is kind of misleading.

Example: let's assume there is plugin logger which has a `printToConsole` API and usage `evaluate` API from taiko.

```javascript
export function printToConsole(message) {
    let options = { args: [message] };
    await taiko.evaluate((element, args) => {
        console.log(args);
    });
    descEvent.emit('success', 'Printed' + message + 'to the browser console.')
}
```

Now when this API gets called the message printed is

```console
> logger.printToConsole("Hello World")
✔ Evaluated given script. Result:undefined
✔ Printed 'Hello world' to the browser console
```
Which is wrong. Instead, it should be just
```console
> logger.printToConsole("Hello World")
✔ Printed 'Hello world' to the browser console
```

### Proposal

Add an `omitDescription` option to the APIs which tell APIs to not print the description.

```javascript
export function printToConsole(message) {
    let options = { omitDescription: true, args: [message] };
    await taiko.evaluate((element, args) => {
        console.log(args);
    }, options);
    descEvent.emit('success', 'Printed' + message + 'to the browser console.')
}
```

MakodeAjay:
Adds omitDescription option in evaluate
Fixes #806

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -2737,6 +2737,9 @@ module.exports.evaluate = async (selector, callback, options = {}) => {
       returnByValue: true,
     });
   });
+  if (options.omitDescription !== true) {
+    descEvent.emit('success', 'Evaluated given script');
+  }
   descEvent.emit('success', 'Evaluated given script');
   return result.result.value;
 };
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -2740,7 +2740,6 @@ module.exports.evaluate = async (selector, callback, options = {}) => {
   if (options.omitDescription !== true) {
     descEvent.emit('success', 'Evaluated given script');
   }
-  descEvent.emit('success', 'Evaluated given script');
   return result.result.value;
 };
 
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -2709,11 +2709,12 @@ module.exports.beforeunload = (callback) => dialog('beforeunload', '', callback)
  * NOTE : In callback, we can access only inline css not the one which are define in css files.
  * @param {Object} options - options.
  * @param {boolean} [options.waitForNavigation=true] - Wait for navigation after the click. Default navigation timeout is 15 seconds, to override pass `{ navigationTimeout: 10000 }` in `options` parameter.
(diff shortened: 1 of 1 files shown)


lib/taiko.js
```diff
@@ -2737,6 +2737,9 @@ module.exports.evaluate = async (selector, callback, options = {}) => {
       returnByValue: true,
     });
   });
+  if (options.omitDescription !== true) {
+    descEvent.emit('success', 'Evaluated given script');
+  }
   descEvent.emit('success', 'Evaluated given script');
```
zabil, line 2743: Is this line needed as there is an if condition just before it?

MakodeAjay: Yes @zabil . So we are validating if the description is already present then don't print an unnecessary message because
all the taiko APIs prints a description of the action they performed.

MakodeAjay: @zabil I removed the below line which is outside the "if" condition.

zabil:
Thanks @MakodeAjay for looking at this.
I took at look at the original issue #806 and although the issue talks about the `evaluate`, which your PR fixes. This has to be extended to all API's not just the `evaluate` API. 

There can be an easier and global way to handle this by not checking the condition and passing this option for every API.

zabil:
@MakodeAjay Had a discussion about this. It's ok to keep this simple and only provide this option for evaluate. I've added a few review comments.

zabil:
@MakodeAjay Just reaching out to ask of you have questions about my earlier comments.

zabil (approved):

PR #1655, opened 2020-11-01
Match input with placeholders

### LINKED ISSUE #1624: Smart selectors fail to find search bar using text placeholder
**Is your feature request related to a problem? Please describe.**
Smart selectors do not work when deal with search bar placeholders using text inside the bar as target
Related to https://github.com/getgauge/taiko/issues/1625

So, if you go to https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_search_placeholder and then run: _write("AAAA", textBox("Filter items"))_ - Taiko won't write into the search bar

Running: _write("AAAA", textBox({**placeholder**: "Filter items"}))_ - it'll work

_In the best case, the functionality already exists but I'm missing something - if so, I'd appreciate some pointers :)
**Describe the solution you'd like**
To be able to interact with elements with placeholders visually available

**Describe alternatives you've considered**
_write("AAAA", textBox("Filter items"))_ should work just like _write("AAAA", textBox({placeholder: "Filter items"}))_

**Additional context**
This is an example of the kind of input with placeholders:
image

saikrishna321:
Fixes #1624


lib/elementWrapper/textBoxWrapper.js
```diff
@@ -21,17 +21,30 @@ function getTextBoxElementWithLabel(searchElement, label) {
...
-      //check label with attribute for
-      const labelForElement = searchElement.getElementById(labelFor);
-      checkAndPushElement(labelForElement);
-    } else {
-      // check child node of label tag
-      matchingLabel.childNodes.forEach((elem) => {
-        checkAndPushElement(elem);
-      });
+
+  const matchingPlaceHolders = [...searchElement.querySelectorAll('input')].filter((labelElem) => {
```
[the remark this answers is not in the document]
zabil: The method can be renamed from `getTextBoxElementWithLabel` to `getTextBoxElementWithLabelOrPlaceholder`

test/unit-tests/textBox.test.js
```diff
@@ -408,7 +408,7 @@ describe(test_name, () => {
                         </div>
                         <div name="withWrappedLabel">
                             <label>
-                                <input type="${inputType.type}"/>
+                            <input type="${inputType.type}" placeholder='Filter'/>
```
zabil, line 411: Can we have a separate input element with only the placeholder for the test?  I notice that this input is used for wrapped label test.

zabil (changes requested):

saikrishna321:
@zabil @NivedhaSenthil ping

zabil:
I added a few comments

test/unit-tests/textBox.test.js
```diff
@@ -723,6 +723,48 @@ describe(test_name, () => {
...
+    let filePath;
+    before(async () => {
+      let innerHtml = `
+      <!DOCTYPE html>
+      <html>
+      <body>
+      
+      <input type="search" id="mySearch" placeholder="Filter items">
+      
+      <p>Click the button to change the placeholder text of the search field.</p>
```
zabil, line 736: Is this section and the `<script>` part required? The tests don't see to use them

zabil:
@saikrishna321 was going to test this. But before that just wanted to check if this will work if the input element is inside a shadow dom. I don't see anything specific to cover that condition.

zabil (approved):
Tested, all is good. Marked this as release candidate. 
Also noticed that the DCO may not be signed on one of the commits.

Issue #1652, opened 2020-11-01
Dropdown multiple values not supported in Typescript


fmpanelli:
**Describe the bug**
Since PR #1602 Taiko supports multiple value selection for dropdowns, but that does not work in Typescript due to type inconsistency (as already mentioned in the PR [comments](https://github.com/getgauge/taiko/pull/1602#issuecomment-714417426))

**To Reproduce**
Steps (or script) to reproduce the behavior:
1. Create a new folder and cd into it
2. Initialize the npm project with `npm init`
3. Add Taiko to the project `npm i taiko`
4. Add Typescript to the project `npm i -D typescript`
5. Create a new file like:
```typescript
//index.ts
import  { dropDown } from "taiko";
async function main() {
  const myDropdown = dropDown('Vehicle:');
  await myDropdown.select(['Car','Van'])
  const values: string[] = await myDropdown.value()
}
```
6. try compiling with `npx tsc`
7. you get an error like:
```
index.ts:6:9 - error TS2322: Type 'string' is not assignable to type 'string[]'.
6   const values: string[] = await myDropdown.value()
```

**Logs**

```
index.ts:6:9 - error TS2322: Type 'string' is not assignable to type 'string[]'.
6   const values: string[] = await myDropdown.value()
```

**Expected behavior**
`DropDownWrapper` now extends `ValueWrapper` which has a method `value()` returning `Promise<string>`.

I think `DropDownWrapper` should extend a new `MultiValueWrapper` which has a method `values()` returning `Promise<string | Array<string> | Array<number>>`.
The `ValueWrapper` should remain as it is for the elements that actually implement that behavior such as `ColorWrapper`, `TextBoxWrapper`, etc.

**Screenshots**

**Versions:**
 - Taiko: 1.0.26
 - OS: Win10 home 10.0.19041
 - Node.js: v14.4.0
 - Typescript: 4.0.5

**Additional context**

BugDiver:
> I think DropDownWrapper should extend a new MultiValueWrapper which has a method values() returning Promise<string | Array<string> | Array<number>>.
The ValueWrapper should remain as it is for the elements that actually implement that behavior such as ColorWrapper, TextBoxWrapper, etc.

@felipdc Since you already have the fix in your mind, would you like to raise a PR for the same?

fmpanelli:
> @fmpanelli Since you already have the fix in your mind, would you like to raise a PR for the same?

Sure! Hopefully I'll be able to submit a PR this evening.

PR #1659, opened 2020-11-03
Making DropDownWrapper extend MultiValueWrapper

fmpanelli:
`DropDownWrapper `previously extended `ValueWrapper `which has a method `value()` returning `Promise<string>`.

With this commit `DropDownWrapper` extends a new `MultiValueWrapper` which has a method `values()` returning `Promise<string | string[] | number[]>`.

The `ValueWrapper` remains as it is for the elements that actually implement that behavior such as `ColorWrapper`, `TextBoxWrapper`, etc.

Related to issue #1652

### DIFF
--- a/lib/elementWrapper/dropDownWrapper.js
+++ b/lib/elementWrapper/dropDownWrapper.js
@@ -1,5 +1,5 @@
 const DropDown = require('../elements/dropDown');
-const MultiValueWrapper = require('./MultiValueWrapper');
+const MultiValueWrapper = require('./multiValueWrapper');
 const { firstElement, getElementGetter } = require('./helper');
 const { $function } = require('../elementSearch');
 
--- a/lib/elementWrapper/dropDownWrapper.js
+++ b/lib/elementWrapper/dropDownWrapper.js
@@ -1,5 +1,5 @@
 const DropDown = require('../elements/dropDown');
-const ValueWrapper = require('./valueWrapper');
+const MultiValueWrapper = require('./MultiValueWrapper');
 const { firstElement, getElementGetter } = require('./helper');
 const { $function } = require('../elementSearch');
 
@@ -32,9 +32,9 @@ function getDropDownElementWithLabel(searchElement, label) {
 /**
  * Behaves the same as ValueWrapper + select().
  * Represents HTML [Select](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select)
- * @extends {ValueWrapper}
+ * @extends {MultiValueWrapper}
  */
-class DropDownWrapper extends ValueWrapper {
+class DropDownWrapper extends MultiValueWrapper {
   constructor(attrValuePairs, _options, ...args) {
     super('DropDown', 'label', attrValuePairs, _options, ...args);
     this._get = getElementGetter(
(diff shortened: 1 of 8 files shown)


zabil:
I see that this pull request changes an existing method `value` to `values`. This breaks backward compatibility for tests that already use `value`.

In the wild, I see more usage of single value dropdown than multi value dropdown. So most of the usage of `values` will probably return only one value. Also, having one method of getting value for a control like `text` makes it easier to discover.

From the tests I see that `values` works for return one value as well an array of values. Can we rename the method to `value`? 

Internally, I think it's ok to use 'multiValueWrapper.js' to get this behaviour.

fmpanelli:
> I see that this pull request changes an existing method `value` to `values`. This breaks backward compatibility for tests that already use `value`.
> 
> In the wild, I see more usage of single value dropdown than multi value dropdown. So most of the usage of `values` will probably return only one value. Also, having one method of getting value for a control like `text` makes it easier to discover.
> 
> From the tests I see that `values` works for return one value as well an array of values. Can we rename the method to `value`?
> 
> Internally, I think it's ok to use 'multiValueWrapper.js' to get this behaviour.

Hi Zabil, thank you for you swift feedback.

That's a very good point. The fundamental fact is that we are changing the external behavior of the `DropDown` element and we are trying to do that as backward-compatible as possible.

I get your approach. It is more backward-compatible than my first proposition but it will still break callers in two cases:
1. if you have a Typescript caller that makes something like the following:
```
async function myDropDownValue(): Promise<string>{
  const myDropdown = dropDown('Vehicle:');
  await myDropdown.select(['Car','Van']);
  return await myDropdown.value(); // type error: Promise<string|string[]|number[]> is not assignable to Promise<string>
}
```
2. if you have some javascript calling `value()` and expecting a string, but actually receiving an array (albeit I am not 100% sure if this is really possible)

Let me propose a third approach which is even more backward-compatible and it is cleaner from the point of view of types.

We may want to have 2 methods in `MultiValueWrapper`:
- `value(): string` which always return 1 value as `string` (if there are more, we may want to return the first one or to throw)
- `values(): string[]` which returns all the values
In this case `MultiValueWrapper` naturally extends `ValueWrapper` making the transition even smoother.

What do you think?

zabil:
> We may want to have 2 methods in MultiValueWrapper:

Yes, we did think about it in the initial PR and I suggested merging the two methods. But I looks like that's a cleaner approach taking typescript into account. So yes, let's do that.

zabil:
One more thing. Won't fixing the signature of the method in the sample you mentioned fix the compilation issue? Do you See an issue upstream when calling `myDropDownValue`? I am assuming it is ok to expect the same type of `value`

```
async function myDropDownValue(): Promise<string|string[]|number[]> {
  const myDropdown = dropDown('Vehicle:');
  await myDropdown.select(['Car','Van']);
  return await myDropdown.value(); // type error: Promise<string|string[]|number[]> is not assignable to Promise<string>
}
```

fmpanelli:
Yes Zabil, changing the calling code will fix it.
My point is that it is a breaking change, meaning that you have to change the client code to make it work properly.
Of course you need to change not only the method `myDropDownValue` but also its callers.

zabil:
> My point is that it is a breaking change

I get it now. It breaks for consumers of the original typescript reference and that shouldn't happen. Two method approach is the best for current javascript and typescript users.

NivedhaSenthil:
@fmpanelli any updates ?

zabil:
Closing this to prevent accidental merges as this is a breaking change. Will re-open when the required changes are pushed.

Issue #1604, opened 2020-10-16
Error The Browser instance was closed


Alenechka:
**Describe the bug**
In the end of every test run of single scenario or full spec I am facing the issue:
Error: The Browser instance was closed either via `closeBrowser()` call, or it exited for reasons unknown to Taiko. You can try launching a fresh instance using `openBrowser()` or inspect the logs for details of the possible crash.
So I am not able to get any report even if test is passed. 

**To Reproduce**

1. Run any spec or scenario using gauge( via VS code plugin or by terminl command gauge run --env default)
For example my steps implementation consists of 

```
beforeSuite(async () => {
        await openBrowser({
    beforeunload(async () => await accept())
});

afterSuite(async () => {
    await closeBrowser();
});

step("Click <text>", async (text) => {
    await click(text, { exactMatch: true });
});

step("Navigate to page <url>", async (url) => {
    await goto(url);
});
```
And spec is:
 
```
# for test

test

## test

* Navigate to page "https://www.google.com/"
* Click "Почта"
```

**Logs**

```
Error: The Browser instance was closed either via `closeBrowser()` call, or it exited for reasons unknown to Taiko. You can try launching a fresh instance using `openBrowser()` or inspect the logs for details of the possible crash.
at errorMessageForBrowserProcessCrash (gauge\node_modules\taiko\lib\browserLauncher.js:298:11)
at ChildProcess.browserExitEventHandler (gauge\node_modules\taiko\lib\browserLauncher.js:312:49)
at Object.onceWrapper (events.js:422:26)
at ChildProcess.emit (events.js:315:20)
at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
Error ----------------------------------

[Gauge]
js runner quit unexpectedly. Check logs for more details.
ErrorMessage: transport is closing

Get Support ----------------------------
	Docs:          https://docs.gauge.org
	Bugs:          https://github.com/getgauge/gauge/issues
	Chat:          https://spectrum.chat/gauge

Your Environment Information -----------
	windows, 1.1.4, c982558
	html-report (4.0.12), java (0.7.10), js (2.3.13), screenshot (0.0.1), spectacle (0.1.4), xml-report (0.2.3)
Error: Tests failed.
```

**Expected behavior**
Test passed or failed but report is generated sucessfully.

**Versions:**
```
 - Gauge version: v1.1.4
 - Taiko v1.0.23 using Chrome(not Chromium)
[...]

PR #1680, opened 2020-11-12
#1604 dont close page on windows headful mode

NivedhaSenthil:
https://bugs.chromium.org/p/chromium/issues/detail?id=1147809

### DIFF
--- a/lib/taiko.js
+++ b/lib/taiko.js
@@ -38,7 +38,6 @@ const {
   setClickOptions,
 } = require('./config');
 const fs = require('fs-extra');
-const os = require('os');
 const path = require('path');
 const childProcess = require('child_process');
 const crypto = require('crypto');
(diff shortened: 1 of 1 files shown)


lib/taiko.js
```diff
@@ -306,18 +307,20 @@ const _closeBrowser = async () => {
     // remove listeners other than JS dialog for beforeUnload on client first to stop executing them when closing
     await _client.removeAllListeners();
     pageHandler.addJavascriptDialogOpeningListener();
-    if (!defaultConfig.firefox) {
+    //TODO: Remove check once fixed https://bugs.chromium.org/p/chromium/issues/detail?id=1147809
+    if (!defaultConfig.firefox && !(os.type().includes('Windows') && defaultConfig.headful)) {
```
zabil, line 311: I think this without having to import a library and having an exact match

https://nodejs.org/api/process.html#process_process_platform

```suggestion
    if (!defaultConfig.firefox && !(process.platform == 'win32' && defaultConfig.headful)) {
```

It also returns win32 for 64 bit systems. The same is available in `os` too.

https://nodejs.org/api/os.html#os_os_platform

PR #1725, opened 2020-12-10
Cleanup connections

NivedhaSenthil:
- pull out connection logic
- move domain logic to respective handlers

### DIFF
--- a/lib/handlers/networkHandler.js
+++ b/lib/handlers/networkHandler.js
@@ -84,9 +84,7 @@ const setUserAgent = async (deviceEmulate) => {
   });
 };
 
-const setCookie = async (options) => {
-  await network.setCookie(options);
-};
+const setCookie = async (options) => await network.setCookie(options);
 
 const getCookies = async (options) => {
   return await network.getCookies(options);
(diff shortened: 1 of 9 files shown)


lib/handlers/networkHandler.js
```diff
@@ -78,7 +78,28 @@ const setNetworkEmulation = async (networkType) => {
...
+const setCookie = async (options) => await network.setCookie(options);
+
+const getCookies = async (options) => {
+  return await network.getCookies(options);
+};
+
+const clearBrowserCookies = async () => await network.clearBrowserCookies();
+
+const deleteCookies = async (options) => await network.deleteCookies(options);
+
```
zabil, line 96: does this belong to network handler or browser handler?

zabil (changes requested):

lib/handlers/networkHandler.js
NivedhaSenthil: they belong to network handler as they are methods of Network domain in CDP ... https://chromedevtools.github.io/devtools-protocol/tot/Network/

zabil (approved):

PR #1752, opened 2020-12-22
Documentation page with side bar (Work in progress)

zabil:
Proposal to modify the existing layout to look something like below. 

Screenshot 2020-12-22 at 16 30 12


zabil:
More changes to make it look like

Screenshot 2020-12-23 at 15 56 11

zabil:
Closing this as after the testing for text based browsers and accessibility the current site design is much easier to navigate and read.

PR #1819, opened 2021-01-26
Fix action on invisible

NivedhaSenthil:
- Remove option `selectHiddenElements` on selectors
- Introduce option `force` on page actions to enforce action irrespective of element being unactionable
- Element wrapper chooses first visible element to perform action if nothing is visible returns the first available element


lib/taiko.js
```diff
@@ -906,6 +906,7 @@ module.exports.title = title;
  * @param {number} [options.elementsToMatch=10] - Number of elements to loop through to match the element with given selector.
  * @param {string[]} [options.waitForEvents = ['firstMeaningfulPaint']] - Events available to wait for ['DOMContentLoaded', 'loadEventFired', 'networkAlmostIdle', 'networkIdle', 'firstPaint', 'firstContentfulPaint', 'firstMeaningfulPaint', 'targetNavigated']
  * @param {number} [options.waitForStart=100] - time to wait for navigation to start. Accepts time in milliseconds.
+ * @param {boolean} [options.force=false] - Set force to true to force doing action irrespective of element being unactionable(invisible, disabled etc)
```
zabil, line 909: Is it ok to shorten the message as follows?
Can we also add example code of passing this option?

```suggestion
 * @param {boolean} [options.force=false] - Set to true to perform action on hidden/disabled elements
```

NivedhaSenthil: Done 👍

zabil (approved):

klhex:
@NivedhaSenthil You removed the definition for SelectionOptions in index.d.ts but currently they are still being referenced in the 'Selectors' section of index.d.ts. Example:
```javascript
// https://docs.taiko.dev/api/image
export function image(
  selector: SearchElement,
  options?: SelectionOptions | RelativeSearchElement,
  ...args: RelativeSearchElement[]
): ImageWrapper;
```

NivedhaSenthil:
Ideally options as a parameter can be removed from the selectors except for ones that are using Text based searches.. but for now it is fine to just remove the `SelectionnOptions` as mentioned above in your PR.

PR #1842, opened 2021-02-03
Adds Typescript support for fix of issue #1771

### LINKED ISSUE #1771: FileField error "is not visible" despite 'selectHiddenElements: true' since Taiko v1.1.1
**Describe the bug**
Since Taiko v1.1.1 (incl. v1.1.2) I get an "is not visible" error when trying to upload a file via fileField with option 'selectHiddenElements: true'. The same code works fine until incl. Taiko v1.1.0. (Maybe a regression caused by https://github.com/getgauge/taiko/pull/1759?)

**To Reproduce**
Steps (or script) to reproduce the behavior:
1. Create an upload container as per the following HTML code:
```html
<div class="c-upload-container">
    <div class="c-upload-container__field u-mb-none">
        <div class="c-upload" tabindex="0">
            <input id="monthly-net-income-upload" class="c-upload__input" accept="image/jpeg,application/pdf"
                multiple="" type="file" autocomplete="off" tabindex="-1" style="display: none" /><span
                class="c-upload__description"><i class="c-upload__icon" aria-hidden="true"></i><span
                    class="c-upload__description-text">Ziehen Sie die hochzuladenden Dateien in diesen
                    Bereich</span><span class="c-upload__description-supplementary-text">oder klicken Sie auf "Dateien
                    auswählen"</span></span><span class="c-upload__actions"><span
                    class="c-upload__button c-btn"><span>Dateien auswählen</span></span></span><span
                class="c-upload__disabled"><i class="c-upload__disabled-icon" aria-hidden="true"></i><span
                    class="c-upload__disabled-text">Sie haben bereits 5 Dateien hochgeladen oder zum Hochladen
                    ausgewählt. Diese können Sie in der Dateiliste verwalten.</span></span>
        </div>
    </div>
    <fieldset name="monthly-net-income-upload-validation" class="o-fieldset u-mb-none">
        <input hidden="" name="monthly-net-income-upload-validation" value="" />
    </fieldset>
</div>
```

Visual representation:
image

2. Execute the following Taiko code on it (use your own .jpg file/path):
```javascript
      await attach(
        'uploads/pirate.jpg',
        fileField(
          { id: 'monthly-net-income-upload' },
          { selectHiddenElements: true }
        )
      );
```
4. See error logs below

**Logs**

```
Error: FileField[id="monthly-net-income-upload"]is not visible
    at /[...]/node_modules/taiko/lib/actions/pageActionChecks.js:178:11
    at waitAndGetActionableElement (node_modules/taiko/lib/actions/pageActionChecks.js:153:3)
    at attach (node_modules/taiko/lib/actions/attach.js:20:19)
    at Object.module.exports.attach (node_modules/taiko/lib/taiko.js:1085:16)
    at Object.module.exports.<computed> [as attach] (node_modules/taiko/lib/taiko.js:2604:14)
```

**Expected behavior**
File is uploaded successfully w/o error message.

**Versions:**
 - Taiko: 1.1.1
 - OS: macOS 11.1 (20C69)
 - Node.js v14.13.1

**Additional context**
Repro case can be provided via mail if necessary.

klhex:
Fixes #1771.
 
(Adds TS support for changes of https://github.com/getgauge/taiko/pull/1819 relating to the new option ```{ force: true }```.)

Please review. I could only test it for function ```attach``` since this is currently my only use case where I need to deal with hidden elements. Also since I'm a JS/TS beginner, I don't know whether my approach for adding TS support is the right one. But it was a low effort approach, so don't worry about deleting the PR if it's crap. 😄 

In this PR I also removed the leftover references to ```SelectOptions```, see [here](https://github.com/getgauge/taiko/pull/1819#issuecomment-772530886).


types/taiko/index.d.ts
```diff
@@ -440,36 +443,54 @@ export function title(): Promise<string>;
 // https://docs.taiko.dev/api/click
 export function click(
   selector: SearchElement | MouseCoordinates,
-  options?: ClickOptions | RelativeSearchElement,
+  options?: ClickOptions | RelativeSearchElement | HiddenOrDisabledElementsOptions,
```
NivedhaSenthil, line 446: new option has to extend the existing options... For example here user can provide click option along with `HiddenOrDisabledElementOptions`

NivedhaSenthil (changes requested):

types/taiko/index.d.ts
klhex: I see, will perform the necessary adjustments.

klhex: I gave it a go with the more verbose option. At least it compiles. ;)

NivedhaSenthil: instead can we create a `ForceOption` and extend it in all other options of actions like 

```
export interface ForceOption {
  force: boolean;
}
export interface ClickOptions extends NavigationOptions, ForceOption {}
```

klhex: Aha! Now I get it, that looks much simpler and cleaner, thanks for the additional hint. Sorry for not seeing this earlier, will likely push a new update this evening (CET).

klhex: Voilà, implemented as suggested. However, I did not apply your suggested approach to the ```NavigationOptions``` since these are not exclusively used for functions relevant to the new ```ForceOption``` (```clear, focus, hover, mouseAction```) but also for other functions such as ```goto, goBack, goForward```. Therefore, I additionally created this option

```javascript
export interface ForcedNavigationOptions extends NavigationOptions {
  force?: boolean;
}
```

and replaced the ```NavigationOptions``` by it for ```clear, focus, hover, mouseAction```.

NivedhaSenthil: huh no my suggestion was not to update navigation option..interface can extend multiple interfaces `,` separated
there can be just `ForceOption` which can be extended in interfaces required

Please refer https://github.com/getgauge/taiko/blob/5595c99694f04ecb9b5e4505e207990121f7eaa7/types/taiko/index.d.ts#L47

klhex: Sorry, what I meant to say was this: if I did what I understood you suggested to NavigationOptions as well, it would look like this, right?
```javascript
export interface NavigationOptions extends BasicNavigationOptions, EventOptions, ForceOption {
  headers?: object;
  waitForStart?: number;
}
```
However, extending the NavigationOptions with ForceOption like above would have the (IMHO negative) side-effect of automatically adding the new ```force: boolean``` option to functions which actually don't support this option (according to the Taiko documentation), such as ```goto(), goBack(), goForward()``` because these functions are also using the NavigationOptions:
```javascript
export function goto(url: string, options?: NavigationOptions): Promise<Response>;
export function goBack(options?: NavigationOptions): Promise<void>;
export function goForward(options?: NavigationOptions): Promise<void>;
```
In order to avoid that, I created
```javascript
export interface ForcedNavigationOptions extends NavigationOptions {
  force?: boolean;
}
```
and added it (by replacing NavigationOptions with ForcedNavigationOptions) to the following functions: ```clear(), focus(), hover(), mouseAction()``` like so:
```javascript
export function focus(selector: SearchElement, options?: ForcedNavigationOptions): Promise<void>;
```
This way, ```clear(), focus(), hover(), mouseAction()``` would now also support the new ```force: boolean``` option (as they should do according to the docs) but ```goto(), goBack(), goForward()``` continue not supporting it (which according to the docs is correct). See what I mean? 😃

NivedhaSenthil: I see where you are coming from... thanks for the explanation 👍

types/taiko/index.d.ts
```diff
@@ -171,6 +171,14 @@ export interface Response extends BasicResponse {
   redirectedResponse?: BasicResponse[];
 }
 
+export interface ForceOption {
+  force?: boolean;
+}
+
+export interface ForcedNavigationOptions extends NavigationOptions {
```
NivedhaSenthil, line 178: Can we update it to 
`export interface ForcedNavigationOptions extends NavigationOptions, ForceOption {}`

NivedhaSenthil:
Please resolve conflicts and update version for a release

types/taiko/index.d.ts
klhex: Smart idea, thanks! Will adjust accordingly and push soon.

NivedhaSenthil (approved):

NivedhaSenthil:
Thanks for the contribution @klhex :)

PR #1955, opened 2021-03-25
Fix evaluate passing data code snippet

snowron:
I add 'args' key to message variable because it didn't work. There was missing part.


sriv (approved):

sriv:
@snowron thanks for the contrib. This repo requires that the commits be signed. Please see details here: https://github.com/getgauge/taiko/pull/1955/checks?check_run_id=2194211090

snowron:
Tried twice to pass the tests but 'FTs - NodeJS 14 & ubuntu-latest' always break the pipeline. Unfortunately dependency bot has same problem.

Issue #1996, opened 2021-04-19
Add custom message to waitFor API


saikrishna321:
Current waitFor api with selector does not allow to pass custom message.

saikrishna321:
Released in v1.2.4

PR #1997, opened 2021-04-19
Custom message for waitFor API

saikrishna321:
Address #1996


lib/taiko.js
```diff
@@ -2323,12 +2323,21 @@ module.exports.into = (value) => value;
...
  * @return {promise}
  */
 
-const waitFor = async (element, time) => {
+const waitFor = async (element, time, options = {}) => {
   validate();
-  let timeout = time || defaultConfig.retryTimeout;
+  let timeout;
+  let message;
+  if (isObject(time)) {
```
NivedhaSenthil, line 2334: `waitFor(5000,{"message":"Timed out"})` seems this case is not covered

NivedhaSenthil: Resolving as we won't encounter this..since wait with only time simply does an explicit time wait

NivedhaSenthil (approved):

PR #2008, opened 2021-04-24
Perform click on given positions

saikrishna321:
This PR opens opportunity to perform click on given element at a certain position. 

Added Positions: left, right, topRight, topLeft, bottomRight, bottomLeft

```
click($('#action-canvas'), {position: 'right'})
click($('#action-canvas'), {position: 'left'})
click($('#action-canvas'), {position: 'bottomRight'})
click($('#action-canvas'), {position: 'bottomLeft'})
click($('#action-canvas'), {position: 'topRight'})
click($('#action-canvas'), {position: 'topLeft'})
```

Screenshot 2021-04-24 at 10 53 22 AM

@NivedhaSenthil @zabil Creating this PR for early feedbacks.

### DIFF
--- a/lib/handlers/domHandler.js
+++ b/lib/handlers/domHandler.js
@@ -34,6 +34,54 @@ async function boundingBoxCenter(e) {
   };
 }
 
+async function boundingBoxRight(e) {
+  const box = await assertBoundingBox(e);
+  return {
+    x: box.x + Math.floor(box.width) - 1,
+    y: box.y + box.height / 2,
+  };
+}
+
+async function boundingBoxTopLeft(e) {
+  const box = await assertBoundingBox(e);
+  return {
+    x: box.x,
+    y: box.y,
+  };
+}
+
+async function boundingBoxTopRight(e) {
+  const box = await assertBoundingBox(e);
+  return {
+    x: box.x + box.width - 1,
+    y: box.y,
+  };
+}
+
+async function boundingBoxBottomRight(e) {
+  const box = await assertBoundingBox(e);
+  return {
+    x: box.x + box.width - 1,
+    y: box.y + box.height - 1,
+  };
+}
+
+async function boundingBoxBottomLeft(e) {
+  const box = await assertBoundingBox(e);
+  return {
+    x: box.x,
+    y: box.y + box.height - 1,
+  };
+}
+
+async function boundingBoxLeft(e) {
+  const box = await assertBoundingBox(e);
+  return {
+    x: box.x,
+    y: box.y + box.height / 2,
+  };
+}
+
 async function boundBox(e) {
   const result = await getBoxModel(e);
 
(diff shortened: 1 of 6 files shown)


zabil:
The API looks fine to me.

lib/taiko.js
```diff
@@ -896,7 +896,8 @@ module.exports.title = title;
  * await click('Get Started', { navigationTimeout: 60000,  force: true })
  * @example
  * await click('Get Started', { navigationTimeout: 60000 }, below('text'))
- *
+ * @example
+ * await click('Get Started', { navigationTimeout: 60000, position: 'right' }, below('text'))
```
zabil, line 900: I noticed that there's nothing added for this new option at 

https://github.com/getgauge/taiko/blob/310097c072d82ba791db7f044d779340d7ed55ca/types/taiko/index.d.ts#L56-L60

lib/handlers/domHandler.js
```diff
@@ -34,6 +34,73 @@ async function boundingBoxCenter(e) {
...
+    case 'topLeft':
+      return await boundingBoxTopLeft(e);
+    case 'bottomRight':
+      return await boundingBoxBottomRight(e);
+    case 'bottomLeft':
+      return await boundingBoxBottomLeft(e);
+    case 'left':
+      return await boundingBoxLeft(e);
+    case 'right':
+      return await boundingBoxRight(e);
```
zabil, line 60: Can we add a default case too?

I assume this will be centre, in case a position has a value that is not null or any of these strings.

zabil (changes requested):

lib/handlers/domHandler.js
```diff
@@ -34,6 +34,73 @@ async function boundingBoxCenter(e) {
...
+async function boundingBoxRight(e) {
+  const box = await assertBoundingBox(e);
+  return {
+    x: box.x + Math.floor(box.width) - 1,
+    y: box.y + box.height / 2,
+  };
+}
+
+async function boundingBox(position = 'center', e) {
+  switch (position) {
```
saikrishna321, line 46: @zabil If no position is given we are defaulting to center

lib/handlers/domHandler.js
saikrishna321: @zabil https://github.com/getgauge/taiko/blob/cb130b5c75dce55e52cf0e444b201cf96ec1bd31/lib/handlers/domHandler.js#L45-L46

zabil: I think this only handles when position is not set.

However someone could do.

```
click({position: "middle"}) // or some other string
```

We could either not click or throw a warning saying 
```
Could not find position 'middle'. Clicking at center position instead.
```

types/taiko/index.d.ts
```diff
@@ -56,6 +56,7 @@ export interface ReloadOptions extends NavigationOptions {
 export interface ClickOptions extends NavigationOptions, ForceOption {
   button?: 'left' | 'right' | 'middle';
   clickCount?: number;
+  position: string;
```
zabil, line 59: Shouldn't this be 
```suggestion
  position?: string;
```

As position is optional?

saikrishna321:
@zabil Fixed all comments. Thanks for taking time and reviewing the PR.

zabil (approved):

PR #2013, opened 2021-04-27
fix(config): Allow navigation and retry timeout to be read from env

shortstacked:
This will allow us to read navigation and retry timeout values from environment variables and configurations. This will allow us to set defaults outside of our test setup

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,8 +1,8 @@
 const defaultConfig = {
-  navigationTimeout: 30000, //Millisecond
+  navigationTimeout: Number(process.env.TAIKO_NAVIGATION_TIMEOUT) || 30000, //Millisecond
   observeTime: 3000, //Millisecond
   retryInterval: 100, //Millisecond
-  retryTimeout: 10000, //Millisecond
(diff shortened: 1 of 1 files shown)


sriv:
Thanks for the change @shortstacked . May I request the below:

- please see if you can add tests for this change: https://github.com/getgauge/taiko/blob/master/test/unit-tests/config.test.js
- please ensure that your commits are signed-off, this will mark DCO as passed - https://github.com/getgauge/taiko/pull/2013/checks?check_run_id=2449093562

shortstacked:
@sriv I've signed off the commit ( I didn't know that was a thing). The problem with adding unit tests is the changes are around process.env, which I'm unable to stub out or mock. Would you have any advice on how to proceed?

I've noticed other env values such as "criConnectionRetries" don't have tests that cover getting values from process.env, so it might be a bit of a rework to make it happen.

sriv:
Yeah, although it's worth doing, I won't fuss over it since we don't have a pattern in place. It is key to have good test isolation and for a PR as simple as this it's probably a bit much!

thanks for the contrib!

Issue #1005, opened 2020-01-07
scrollTo() behavior


moogleberry:
### Expected behavior
`scrollTo()` should provide the ability to center the view on the element provided by leveraging underlying features.  

Specifically, using `element.scrollIntoView()` instead of the Chrome-only `element.scrollIntoViewIfNeeded()` internally should provide better future support since `scrollIntoView` is in a Working Draft and `scrollIntoViewIfNeeded` is not on any standards track.  

Further, `scrollIntoView` provides more positioning options such as `{ block: "center" }`, which can be made available by passing a configuration Object through the `scrollTo()` call into the internal call.  These options are important for sites with sticky or floating elements.  Scrolling barely into "view" as the browser sees it may not scroll an element into an uncovered position.  

`scrollIntoViewIfNeeded()` only provides a center on element Boolean, but I have observed this flag to not work on the listed Chromium version below.

### Actual behavior
`scrollTo()` only allows for scrolling barely into view because of the behavior of the internally used `scrollIntoViewIfNeeded()`, and both its lack of positioning options, and Taiko's provision of a pass-through to them. 

### Steps to reproduce
1. openBrowser()
2. goto("https://taiko.gauge.org/#/")
3. scrollTo("This launches the Taiko prompt")

### Versions
```
Taiko 1.0.3
Chromium Version 79.0.3945.88 (Developer Build) Fedora Project (64-bit)
Chromium Version 81.0.3994.0 (Developer Build) (64-bit)
```
[scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)
[scrollIntoViewIfNeeded](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded)

negiDharmendra:
@moogleberry Can you please elaborate your use case about using `scrollIntoView` instead of `scrollIntoViewIfNeeded`. Just want to understand where and how [scrollIntoViewOptions](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) are adding values as far as automation testing is concern.

NivedhaSenthil:
Thanks for bring this up, scrollTo is modified to `scrollIntoView` in favour of supporting firefox currently.

Issue #1565, opened 2020-10-02
Click on navigation bar item does not trigger navigation (scroll down)


klhex:
**Describe the bug**
When a navigation item on a page with a navigation bar is clicked then the scrolling down to the page's section referenced by the navigation item does only start for a short moment but then stops. It only succeeds fully when the same navigation item is clicked again. 

The next issue concerns clicking on 'MEHR ERFAHREN' once the downwards navigation has succeeded (currently, by clicking again on 'Auto Abo abschließen'): If one tries to click on the leftest 'MEHR ERFAHREN' button like as follows

```javascript
await click('MEHR ERFAHREN', {}, below('Premium-Paket'));
```
then the button is scrolled into view and you get this error: "Error: Element matching text "MEHR ERFAHREN" is covered by other element".

If one adds `await hover('MEHR ERFAHREN');` before the above mentioned click, then the click succeeds, no error message.

**To Reproduce**
```javascript
const { openBrowser, goto, click, closeBrowser } = require('taiko');
(async () => {
    try {
        await openBrowser({headless: false});
        await goto('vwfs.de/autoabo.html');
        await click('Ja, ich willige ein');
        await click('Auto Abo abschließen');
    } catch (error) {
        console.error(error);
    } finally {
        await closeBrowser();
    }
})();
```

**Expected behavior**
1st issue expected behavior: After the click on 'Auto Abo abschließen' the page should be scrolled down to where the navigation item is referencing to.

2nd issue expected behavior: The above mentioned click on 'MEHR ERFAHREN' should lead for the right element to be found, scrolled into view and be clicked instead of throwing the above mentioned error message.

**Versions:**
 - Taiko: 1.0.23
 - MacOS Catalina: 10.15.6 (19G2021)
 - Node.js: v14.7.0

**Additional context**
It appears that the scrolling issue is related to the size of the browser window. Running the same test with a lower browser window height, for example, seems to make the scrolling (after the click on 'Auto Abo abschließen') work. But it still seems to be relatively brittle.

NivedhaSenthil:
I could that when we try to click `Auto Abo abschließen` for the first time the page is scrolled once the coordinates to click is calculated and hence Taiko is clicking on wrong coordinates. Seems like highlighting element on action is causing the scroll, trying to find why that is happening.

PR #2066, opened 2021-05-22
Change scrollTo action behavior, referal about taiko issue #1005

dorion:
There are a lot of items on my page like wizards that are sticky footers, and when I scroll to an item that's underneath it, the item always stays underneath it.


NivedhaSenthil:
I feel it is better to expose the scroll to position as an option ..since we have seen different behaviors needed for particular cases.

Previously the position was altered because this issue https://github.com/getgauge/taiko/issues/1565

NivedhaSenthil:
@dorion would you like to take up on adding an option to alter scroll position.. can help if you need any pointers

lib/actions/scrollTo.js
```diff
@@ -4,10 +4,10 @@ const { defaultConfig, setNavigationOptions } = require('../config');
 const { highlightElement } = require('../elements/elementHelper');
 const { doActionAwaitingNavigation } = require('../doActionAwaitingNavigation');
 
-const scrollToElement = async (element) => {
+const scrollToElement = async (element, alignments) => {
```
zabil, line 7: I think it's better to call this alignment

```suggestion
const scrollToElement = async (element, alignment) => {
```

lib/actions/scrollTo.js
```diff
@@ -4,10 +4,10 @@ const { defaultConfig, setNavigationOptions } = require('../config');
...
   function scrollToNode() {
     const element = this.nodeType === Node.TEXT_NODE ? this.parentElement : this;
-    element.scrollIntoView({ block: 'nearest', inline: 'nearest' });
+    element.scrollIntoView({ block: alignments.block, inline: alignments.inline });
```
zabil, line 10: Does it make sense to make this as follows?

```suggestion
    element.scrollIntoView(alignment);
```

As mentioned here https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

alignment has other options too

lib/taiko.js
```diff
@@ -1255,16 +1255,23 @@ async function mouseAction(selector, action, coordinates, options = {}) {
...
- *
+ * @param {number} [options.navigationTimeout=30000] - Navigation timeout value in milliseconds for navigation after click.
+ * @param {number} [options.waitForStart = 100] - time to wait for navigation to start. Accepts value in milliseconds.
+ * @param {Object} options.alignments
+ * @param {string} [options.alignments.block = 'nearest'] - Defines vertical alignmen.
+ * @param {string} [options.alignments.inline = 'nearest'] - Defines horizontal alignmen.
+ * 
  * @returns {Promise<void>}
  */
 module.exports.scrollTo = async (selector, options = {}) => {
```
zabil, line 1268: It will be good to add an example on how to pass the options and link to
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView for reference.

zabil (changes requested):
Thanks for picking this up! Added a few comments.

lib/taiko.js
dorion: Thank you for your review, I tied resolve and accepting your mentions.

lib/taiko.js
```diff
@@ -1247,24 +1247,36 @@ async function mouseAction(selector, action, coordinates, options = {}) {
...
+ * The alignment parameters can be overridden, see below.
+ * Tthe possible values reference are available at the https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView.
+ * 
  *
  * @example
  * await scrollTo('Get Started')
  * @example
  * await scrollTo(link('Get Started'))
+ * @example
+ * await scrollTo('Get Started',{ aligments: { block: 'center', inline: 'center' }})
```
dorion, line 1259:
```suggestion
 * await scrollTo('Get Started',{ alignments: { block: 'center', inline: 'center' }})
```

dorion:
I changed the code to test will be great(green) again.

dorion:
I fixed DCO too

dorion:
Can I do anything else? Missing something?

NivedhaSenthil (approved):

Issue #2149, opened 2021-06-25
Error while drag n drop - ProtocolError: Invalid parameters


sk1981:
**Describe the bug**
When I try to use the drag-n-drop api the following error is thrown: `ProtocolError: Invalid parameters (Failed to deserialize params.force - BINDINGS: double value expected at position 13)`

**To Reproduce**
Steps (or script) to reproduce the behavior:
1. In any step, for a non draggable element
2. use the dragndrop function
```
await dragAndDrop(
    elem,
    { up: 0, down: 0, left: 0, right: 600 },
    { force: true }
  );
```

3. Running the test gives the error above

**Logs**

```
 ✔ Browser closed
ProtocolError: Invalid parameters (Failed to deserialize params.force - BINDINGS: double value expected at position 13)
    at ../node_modules/chrome-remote-interface/lib/chrome.js:94:35
    at Chrome._handleMessage (../node_modules/chrome-remote-interface/lib/chrome.js:257:17)
 ....  {
  request: {
    method: 'Input.dispatchMouseEvent',
    params: {
      force: true,
      waitForNavigation: true,
      waitForEvents: [],
      navigationTimeout: 30000,
      waitForStart: 100,
      x: 193.125,
      y: 468.01,
      button: 'left',
      clickCount: 1,
      elementsToMatch: 10,
      type: 'mouseMoved'
    },
    sessionId: undefined
  },
  response: {
    code: -32602,
    message: 'Invalid parameters',
    data: 'Failed to deserialize params.force - BINDINGS: double value expected at position 13'
  }
}

```

**Expected behavior**

**Screenshots**

**Versions:**
 - Taiko: Version: 1.2.5 
 - Chromium: 88.0.4314.0 RELEASE
 - Mac 10.15.6 
 - Node.js v14.15.3
 - 

**Additional context**

I believe the issue is the `force` parameter that is is passed to chromium.  As per [taiko docs](https://docs.taiko.dev/api/draganddrop/)  the value should be boolean:

<img width="493" alt="Screen Shot 2021-06-25 at 12 32 33 PM" src="https://user-images.githubusercontent.com/10632239/123412050-7099fe00-d5b1-11eb-9881-f7d01b0053cb.png">

As per the doc site: https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchMouseEvent

[...]

PR #2168, opened 2021-07-01
Error while drag n drop - ProtocolError: Invalid parameters #2149

Siddharth-Tumre:
Refactored to not conflict the force attribute of Taiko with 'dispatchMouseEvent' from 'Chrome DevTools Protocol'.
And this will throw an error if we try to perform dragAndDrop with 'force' set as true on hidden elements stating "Taiko cannot drag hidden elements".


NivedhaSenthil:
Please fix the failing test case

Siddharth-Tumre:
> Please fix the failing test case

Two of the functional tests were failing. I have fixed those.

lib/actions/dragAndDrop.js
```diff
@@ -15,6 +15,14 @@ const dragAndDrop = async (source, destination, options) => {
     isSelector(destination) || isString(destination) || isElement(destination)
       ? await waitAndGetActionableElement(destination, options.force)
       : destination;
+  if (!(await sourceElem.isVisible())) {
+    throw new Error('Taiko cannot drag hidden elements');
+  }
+  if (isSelector(dest) || isString(dest) || isElement(dest)) {
```
NivedhaSenthil, line 21: isVisible is not applicable of string its enough if we just check isElement since we are already fetching destination element for other cases

Siddharth-Tumre:
I hope everything is good now. @NivedhaSenthil

NivedhaSenthil (approved):

PR #2229, opened 2021-08-02
Cache proxomity element Rect

saikrishna321:
Optimize proximity selector search by caching the calculated bounding rects of given proximity elements.

For Ex: 

```
$('.class-multiple-elements', toRightOf(text('Hello')))
```

`.class-multiple-elements',` returns 200 elements and `toRightOf` was calculated for ever element which impacted the performance.


zabil:
@saikrishna321 can you add a small description of what this change does?

PR #2307, opened 2021-09-09
Remove Firefox support

zabil:
Remove all the branching code related to Firefox because
there are no tests and there are few workaround to support
Firefox.

The aim of this change to make the code base cleaner and avoid issues.
Ideally, it should work without branching code as far as possible.


saikrishna321:
@zabil Until FF supports full CDP. We can live with this. This would remove FF support completely. I'm one of the users on FF

zabil:
Oh ok. Closing this then. I assume that there is some form of testing and compatibility checks.

PR #2340, opened 2021-10-01
Update chromium to revision 925089

gaugebot:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


zabil:
@saikrishna321 there is a valid issue with chromium on windows and we need to fix the failure. Otherwise it will break existing tests on windows because of the proximity selectors will fetch the wrong elements.

I recommend, putting the windows build back.

Issue #2173, opened 2021-07-02
Error while handling range element in react app


arijit91:
**Describe the bug**
While selecting a range value in a react app, it appears that the event handlers are not being triggered correctly

**To Reproduce**
Deployed a sample app to demonstrate the behaviour https://cocky-noether-e87ec6.netlify.app/

Sign up button is enabled only if the age threshold is met. Taiko is able to set the age correctly for the range but the button does not get enabled. It works for a manual run.

I think a similar issue affects timefield and checkbox element types as well

```
const { openBrowser, goto, range } = require('taiko');
> openBrowser()
 ✔ Browser opened
> goto("https://cocky-noether-e87ec6.netlify.app/")
 ✔ Navigated to URL https://cocky-noether-e87ec6.netlify.app/
> range("Age").select(50)
 ✔ Selected value 50 for the given input value 50
```

**Expected behavior**
Button should be enabled for the above set of commands.

**Screenshots**
Manual run:
<img width="190" alt="Screenshot 2021-07-02 at 3 52 00 PM" src="https://user-images.githubusercontent.com/862296/124261362-51b9df80-db4e-11eb-8048-d53225f510be.png">

Taiko run:
<img width="187" alt="Screenshot 2021-07-02 at 3 52 09 PM" src="https://user-images.githubusercontent.com/862296/124261372-541c3980-db4e-11eb-8944-4a13c378d3e9.png">

**Additional context**
Maybe related to https://coryrylan.com/blog/trigger-input-updates-with-react-controlled-inputs

NivedhaSenthil:
Thanks for reporting with a sample, managed to replicate will take a look

aishwarya-g-thoughtworks:
There is a workaround for this issue. Please use the following steps:

```
openBrowser()
goto("https://cocky-noether-e87ec6.netlify.app/")
mouseAction($("#range1"),'press', {x:100,y:10})
```

You can change the values of x and y as appropriate to the application.

PR #2377, opened 2021-10-27
Bug(checkbox|radio): Set checked value on native element setter so that events are triggered properly

lethaldose:
- Fixes: #2173  
- for `radioButton` and `checkBox` in react app
- setting native checked value causes events to be raised correctly by react

Signed-off-by: atulv <atulverma82@gmail.com>

### DIFF
--- a/lib/elements/checkBox.js
+++ b/lib/elements/checkBox.js
@@ -19,6 +19,7 @@ class CheckBox extends Element {
     return result.value;
   }
   async check() {
+    await this.registerNativeValueSetter();
     const navigationOptions = setNavigationOptions({});
     await doActionAwaitingNavigation(navigationOptions, async () => {
       const objectId = this.get();
@@ -49,7 +50,7 @@ class CheckBox extends Element {
 }
 
 function setChecked(value) {
-  this.checked = value;
+  this.setNativeValue(this, 'checked', value);
   ['change', 'input', 'click'].forEach((ev) => {
     let event = new Event(ev, { bubbles: true });
     try {
--- a/lib/elements/element.js
+++ b/lib/elements/element.js
@@ -1,8 +1,8 @@
-const { setNativeValue } = require('./elementHelper');
 /**
  * Abstract Element present on the web page. Extra methods are available based on the element type.
  * @see {ElementWrapper} for methods available
  */
+
 class Element {
   constructor(objectId, description, runtimeHandler) {
     this.objectId = objectId;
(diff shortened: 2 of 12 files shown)


sriv:
@NivedhaSenthil would you be able to look at this once it is ready? Thanks!

zabil:
> this update is now defining setNativeValue() on the element prototype itself i.e HTMLInputElement in the case of range, radiobutton and checkbox. So irrespective of iframe the function is found while doing runtime evaluation

I think this is an acceptable solution.

> Another option is to keep executionContextId along with objectId for each element. While calling runtimeHandler.runtimeEvaluate we can then pass correct executionContetId for the iframe in which element exists. 

Prefer a stateless approach without having to store these, IMHO.

lethaldose:
i have fixed the failing functional tests here - https://github.com/getgauge/taiko/pull/2377/commits/7c58d559a06041546ad8fdf1a93cc04637c25777

They were failing and passing randomly, so made them more deterministic without changing intent of test.
These 2 were failing as browser was not focussing on the correct tab. One was failing as the new tab opened by clicking a link was not focussed. The other was failing as on closing a tab,  the next tab to get focussed was not as expected.
Tab focus is controlled by users browser settings.

NivedhaSenthil:
All functions injected to browser are appended with `taiko_` to avoid any name collision, please refer https://github.com/getgauge/taiko/blob/f33c9f45bbc1d0f6e38b50f4540f38cea79adfb0/lib/handlers/runtimeHandler.js#L146... taiko assume that the function is already injected and injects them only if not found into each of the execution contexts. We could try taking leverage of the same.
ObjectID in turn binds to execution contexts IIRC

lethaldose:
> All functions injected to browser are appended with `taiko_` to avoid any name collision, please refer https://github.com/getgauge/taiko/blob/f33c9f45bbc1d0f6e38b50f4540f38cea79adfb0/lib/handlers/runtimeHandler.js#L146
> ... taiko assume that the function is already injected and injects them only if not found into each of the execution contexts. We could try taking leverage of the same.


@NivedhaSenthil function are only injected if the `executionContextId` is passed in as an argument.
`executionContextID` is not available at the element object level, while calling `runtimeCallFunctionOn`. So the function is not injected in the IFrame. hence i tried the approach of binding function to element prototype

NivedhaSenthil:
Huh I see.. ya currently we are caching only the functions executed with executionContextId 
We might have to find a better way to inject the functions.. probably as a listener on context created events. Will create a new issue for that. This looks good so far then...

NivedhaSenthil (approved):

PR #2380, opened 2021-11-01
Added ability to pass useHostName and secure params via openBrowser API 

### LINKED ISSUE #2378: Ability to pass useHostName and secure parameters to chrome-remote-interface via taiko
**Is your feature request related to a problem? Please describe.**
I am trying to run taiko tests on a remote machine. Taiko uses chrome-remote-interface which exposes the parameter `useHostName` to make HTTP calls with hostname rather than an IP address ([code](https://github.com/cyrus-and/chrome-remote-interface/blob/master/lib/external-request.js#L14-L23)). 
Making a request with an IP address to a server behind an NGINX proxy would give the following error:

```html
<html>

<head>
	<title>400 The plain HTTP request was sent to HTTPS port</title>
</head>

<body>
	<center>
		<h1>400 Bad Request</h1>
	</center>
	<center>The plain HTTP request was sent to HTTPS port</center>
</body>

</html>
```

**Describe the solution you'd like**
To prevent this, we can expose the parameter `useHostName` via, maybe, the `openBrowser()` API and handle it pass it accordingly to chrome-remote-interface. Code links to CRI call from taiko([code](https://github.com/getgauge/taiko/blob/master/lib/handlers/targetHandler.js#L26-L31), [code](https://github.com/getgauge/taiko/blob/master/lib/handlers/targetHandler.js#L191-L197))

https://github.com/getgauge/taiko/blob/f33c9f45bbc1d0f6e38b50f4540f38cea79adfb0/lib/handlers/targetHandler.js#L26-L32

**Edit 1**: Taiko's `openBrowser` API also does not support passing the `secure` parameter [docs](https://github.com/cyrus-and/chrome-remote-interface#cdpoptions-callback) to chrome-remote-interface. Thus, it can also be passed via the `openBrowser` API.

Thoughts?
@NivedhaSenthil , @zabil , @saikrishna321

ankur-lt:
The changes deals with the ability to pass `useHostName` and `secure` params via `openBrowser` API.
- Added `useHostName` and `secure` in the `openBrowser` documentation
- Set default value for `useHostName` and `secure` as false to retain taiko's existing behaviour
- Passing the parameter `useHostName` and `secure` in all functions across the library which is initialising chrome-remote-interface
- Fixed the unit tests asserting on the default behaviour of taiko.
- Fixes #2378


lib/taiko.js
```diff
@@ -89,6 +90,7 @@ module.exports.openBrowser = async (
   if ((options.host && options.port) || options.target) {
```
saikrishna321, line 90: @ankur-lt `useHostName` will be set only when host, port or target is set. Is this expected or only `useHostName` with openBrowser is good.

ankur-lt: Yes, this is expected. It will be used only when host, port or target is specified.

zabil:
@saikrishna321 @ankur-lt marking this as a release candidate pending review and merge.

ankur-lt:
@zabil , updated the version to `1.2.8`.

NivedhaSenthil:
@ankur-lt can you please fix the unit tests

ankur-lt:
@NivedhaSenthil , fixed the unit tests and updated the PR to include the `secure` param too.

NivedhaSenthil (approved):

PR #2454, opened 2022-01-09
bump chrome to latest

sriv:


zabil:
It's the same tests failing for windows because the there's a difference in proximity selector logic

```
  1137 passing (57s)
  18 pending
  3 failing

  1) isDisabled
       textBox in form isDisabled() when attribute is set:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true

      at Context.<anonymous> (test\unit-tests\isDisabled.test.js:62:70)

  2) isDisabled
       textBox in form isDisabled() when attribute is not set:

      AssertionError: expected true to be false
      + expected - actual

      -true
      +false

      at Context.<anonymous> (test\unit-tests\isDisabled.test.js:66:71)

  3) within
       test selector with within
         element within should exists:

      AssertionError: expected false to be true
      + expected - actual

      -false
      +true

      at Context.<anonymous> (test\unit-tests\proximitySelectors\within.test.js:75:69)
```

I am able to replicated this on windows (using parallels). Taking a look at it.

sriv:
I saw this too. The problem was that the incorrect textbox was being located. I am not sure why this would only happen on latest chromium, on windows and with node 17 alone.

zabil:
For me this fails on node 14, 16 on windows too. There's definitely a few changes in later versions of chromium for windows. Especially at

https://github.com/getgauge/taiko/blob/e40d18fb9297a1859e65f825137158501b80a451/lib/handlers/domHandler.js#L136-L154

Which returns a value offset by 23 (on later versions of chromium) which then selects the wrong element according to out proximity selector logic.

zabil:
@NivedhaSenthil and I paired on this.

So it looks like on Windows CDP's `getBoxModel` is return values different from other mac and linux. The values aren't different when we call `getContentQuads` so modifying domhandler to use that instead.

cc/ @saikrishna321

sriv:
looks like there are some breaking changes in later versions of chromium

zabil:
It looks like something to do with this option 

https://github.com/getgauge/taiko/blob/e40d18fb9297a1859e65f825137158501b80a451/lib/browserLauncher.js#L261

On the taiko's current version of chrome if i remove `site-per-process` The test fails with the same error while handling iframes. 

In the newer versions of chrome it seems to not disable this flag. Maybe there were changes made around this.

zabil:
I just noticed a pattern in the current test failures. They fail when "TAIKO_HIGHLIGHT_ON_ACTION=false"

For example

```
➜  functional-tests git:(chrome-update) headless=false gauge run specs/pageActionsIFrame.spec
Missing baseUrl in compilerOptions. tsconfig-paths will be skipped
# Page Actions in IFrame
  ## Write, Press, Click	 ✔ ✔ ✔ ✔ ✔ ✔
  ## Double Click	 ✔ ✔ ✔ ✔
  ## Right Click	 ✔ ✔ ✔
  ## Hover	 ✔ ✔ ✔
  ## Drag and drop	 ✔ ✔ ✔

Successfully generated html-report to => /Users/zabilcm/gauge/taiko/test/functional-tests/reports/html-report/index.html

Specifications:	1 executed	1 passed	0 failed	0 skipped
Scenarios:	5 executed	5 passed	0 failed	0 skipped

Total time taken: 16.247s
Updates are available. Run `gauge update -c` for more info.
➜  functional-tests git:(chrome-update) TAIKO_HIGHLIGHT_ON_ACTION=false headless=false gauge run specs/pageActionsIFrame.spec
Missing baseUrl in compilerOptions. tsconfig-paths will be skipped
# Page Actions in IFrame
  ## Write, Press, Click	 ✔ ✔ ✔ ✘
        Failed Step: Get value "Gopher" of Text Box "Username"
        Specification: specs/pageActionsIFrame.spec:8
        Error Message: '' == 'Gopher'
        Stacktrace:
        AssertionError [ERR_ASSERTION]: '' == 'Gopher'
            at HtmlElementAPI.<anonymous> (/Users/zabilcm/gauge/taiko/test/functional-tests/tests/htmlElementAPI.ts:113:12)
            at Generator.next (<anonymous>)
            at fulfilled (/Users/zabilcm/gauge/taiko/test/functional-tests/tests/htmlElementAPI.ts:14:58)
  ## Double Click	 ✔ ✔ ✔ ✔
  ## Right Click	 ✔ ✔ ✘
        Failed Step: Click "Share On Facebook"
        Specification: specs/pageActionsIFrame.spec:30
        Error Message: Element matching text "Share On Facebook" is covered by other element
        Stacktrace:
        Error: Element matching text "Share On Facebook" is covered by other element
            at /Users/zabilcm/gauge/taiko/lib/actions/pageActionChecks.js:189:11
            at waitAndGetActionableElement (/Users/zabilcm/gauge/taiko/lib/actions/pageActionChecks.js:156:3)
            at click (/Users/zabilcm/gauge/taiko/lib/actions/click.js:57:21)
            at module.exports.click (/Users/zabilcm/gauge/taiko/lib/taiko.js:923:16)
            at module.exports.<computed> (/Users/zabilcm/gauge/taiko/lib/taiko.js:2626:14)
  ## Hover	 ✔ ✔ ✔
  ## Drag and drop	 ✔ ✔ ✔

Successfully generated html-report to => /Users/zabilcm/gauge/taiko/test/functional-tests/reports/html-report/index.html

Specifications:	1 executed	0 passed	1 failed	0 skipped
Scenarios:	5 executed	3 passed	2 failed	0 skipped

Total time taken: 17.429s
Updates are available. Run `gauge update -c` for more info.
```

I assume that highlight is doing some kind of focus. Checking why this is happening.

zabil:
@sriv @NivedhaSenthil 

After an epic battle looks like there is green build. This PR now has a bunch of changes mainly around the following.

* Use `getContentQuads` instead of `getBoxModel` because `getBoxModel` does not fetch the right co-ordinates for windows
* Use `document.title` for getting document title because the browser crashes occasionally when a query selector is used for getting titles.
* Navigate to local test files using a locally launched servers are there are changes introduced in the way chrome handles cross site iframes in headless mode.
* Add a wait on highlight in headless mode as the tests were failing when `HIGHLIGHT_ON_ACTION` was set to false.
* Run the tests on node 16 and node 17 is unstable. We'll need to add node 18 support whenever it comes out instead.
* Launch the browser before every scenario instead of relying on incognito windows. Not sure if this is required but this is hoping that the tests will be stable and less prone to crashes

I am not sure if we want to release this yet. Maybe @sguptatw could run test on the bahmni suite of `master` once this is merged to check if everything is ok.

lib/elements/elementHelper.js
```diff
@@ -4,9 +4,12 @@ const overlayHandler = require('../handlers/overlayHandler');
 const { wait } = require('../helper');
 
 const highlightElement = async (element) => {
+  // Adding a wait as tests fail on newer versions of chrome
   if (defaultConfig.highlightOnAction.toLowerCase() !== 'true') {
-    return;
+    await wait(1000);
```
NivedhaSenthil, line 9: Adding an arbitrary wait of 1s before every action would drastically increase the test execution time, I feel this is hiding some other issue behind it. I would like to investigate a little bit on this

NivedhaSenthil:
This is a huge step, really appreciate your perseverance  :)

lib/elements/elementHelper.js
zabil: Sure, but please note the wait would've been there by default as highlight on action is true when running on headful mode.

NivedhaSenthil: ya we have a wait in headful mode because the highlight on the element of action has to be visible for user. We should not be really needing that in headless mode, hope we should be able to find the issue that is needing this wait.

zabil: Absolutely another note. This code block is only called in headful mode

```
    if (defaultConfig.headful) {
      await highlightElement(element);
    }
```

So the wait is not there on headless mode. This is only there on headful mode.

NivedhaSenthil: oops my bad.. some how I was thinking the wait is applicable to headless mode too here if only for headful mode I feel this should be fine

NivedhaSenthil (approved):

PR #2512, opened 2022-02-10
chore: add openBrowser debug logs

louislebrault:


zabil:
@louislebrault Can you give some context on why you need this log message only on the `openBrowser` method? As of now none of the api's write to the log. This is to reduce the verbosity and let the test runner handle it. If you explain your use case maybe it could be a `console.debug` instead of a `console.log`

PR #2537, opened 2022-03-02
Remove div height

zabil:
Tests on github action seem to fail while
asserting on the element as it might be
outside the viewport.

Also adding ts types to get rid of editor warning of missing types.


zabil:
Examples of random test failure
https://github.com/getgauge/taiko/runs/5377546902?check_suite_focus=true

PR #2558, opened 2022-03-17
#2508 fix click api to take co-ordinates as options

saikrishna321:


NivedhaSenthil:
Can we also add a test for this case ?

Issue #2397, opened 2021-11-17
closeBrowser throws error after closeTab when run Headless.


DrRuhe:
**Describe the bug**
using openTab and closeTab leads to closeBrowser throwing errors. 

BUT ONLY WHEN RUN IN HEADLESS.

**To Reproduce**
```
import { closeBrowser, openBrowser, closeTab, openTab } from 'taiko';

async function test() {
    await openBrowser({headless: true});
    await openTab();
    await closeTab();
    await closeBrowser(); // error occurs here 
}

test();
```

**Logs**

```
Error: WebSocket is not open: readyState 2 (CLOSING)
    at sendAfterClose (/home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/ws/lib/websocket.js:903:17)
    at WebSocket.send (/home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/ws/lib/websocket.js:405:7)
    at Chrome._enqueueCommand (/home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/chrome-remote-interface/lib/chrome.js:286:18)
    at /home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/chrome-remote-interface/lib/chrome.js:88:22
    at new Promise (<anonymous>)
    at Chrome.send (/home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/chrome-remote-interface/lib/chrome.js:87:20)
    at handler (/home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/chrome-remote-interface/lib/api.js:33:23)
    at /home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/taiko/lib/connection.js:31:33
    at new Promise (<anonymous>)
    at Proxy.<anonymous> (/home/jakobgerhardt/Projects/ZK2.6/zk-fe-guitest/node_modules/taiko/lib/connection.js:22:24) 
```

**Expected behavior**
No crash 

**Screenshots**

**Versions:**
 - Taiko: Version: 1.2.7 (Chromium: 88.0.4314.0) RELEASE
 - OS Pop!_OS 20.04 LTS
 - Node.js v16.11.0

```
Gauge version: 1.4.1
Commit Hash: a82478a

Plugins
-------
html-report (4.1.2)
java (0.7.15)
js (2.3.17)
[...]

PR #2588, opened 2022-05-16
Add wait to flush interceptors #2397

zabil:
Adding a small wait before closing browser to handle #2397


zabil:
Also notice this fix seems to have fixed the flakiness in tests which was failing earlier at after Scenario `closeBrowser`

PR #2663, opened 2022-11-29
Fix some vulnerabilities and other enhancements

marques-work:
This looks like changes to some random files (and, to be fair, it is). But there's a lot of context in the commit messages.

In summary, this basically does:

1. Upgrade `documentation@14.0.0`
2. Fix a timezone-flaky test
3. Document all necessary browser dependencies in the github actions build; on vanilla ubuntu, the list was insufficient
4. Make linting part of the build and fix lint errors

@NivedhaSenthil could I get a review when you have a moment? I will leave some comments inline to the code for context.


lib/data/devices.js
```diff
@@ -523,54 +523,6 @@ let DEVICES = {
       isLandscape: true,
     },
   },
-  'iPhone 11': {
```
marques-work, line 526: iPhone 11 entries were duplicated. This was picked up by `eslint`. Here, I'm deleting the first instances, lexically, so this should be equivalent as JS would have honored the last instance anyway.

test/unit-tests/timeField.test.js
```diff
@@ -31,13 +31,25 @@ describe(test_name, () => {
...
+  };
+
   var inputTypes = [
     {
       type: 'date',
       name: 'inputType-date',
       min: '2018-01-01',
       max: '2018-12-31',
-      testValue: new Date('2018-09-12'),
+      testValue: localDate(2018, 9, 12),
```
marques-work, line 52: This test fails late night in California. Likely has to do with how `input[type=date[` handles construction from a string vs just passing in an ISO datestring to the Date constructor. I bet the constructor parses the ISO date as midnight UTC.

NivedhaSenthil:
Thanks for taking time to fix this ..I will start reviewing :)

zabil:
@marques-work it is a problem with the node version. I bumped the version to node 18 and got the build to work.

marques-work:
Thanks @zabil!

marques-work:
Whoever sees this next, I think the tests are waiting on approval to run. No rush on it, just don't want it to be missed entirely.

NivedhaSenthil (approved):

NivedhaSenthil:
LGTM

PR #2666, opened 2022-12-03
Canary update all JS dependencies and update chromium to match

marques-work:
- Updates JS dependencies to the latest releases
- Update Chromium to the latest common revision (`branch_base_position`)
- Include script to easily perform canary updates
- Move `updateChromium.js` to the `scripts` folder


marques-work:
@NivedhaSenthil Whenever you get a chance, could you review this PR?

cc: @zabil

BTW, is there any reason we couldn't support ARM64 chromium, at least for `mac_arm` platform? M1 macs are becoming more commonplace and it would be snappier to run without Rosetta.

I'd make a separate PR for that, but wanted to double check in case there was some technical reason it wasn't supported. I don't know if there are any other platforms that really need ARM64 -- prob the most common linux on ARM64 is the raspberry PI 4? I don't know the quality (or exact architecture) of the `Linux_ARM_Cross-Compile` or `Arm` snapshots (are they arm64 or arm32?).

zabil:
> I'd make a separate PR for that, but wanted to double check in case there was some technical reason it wasn't supported.

I don't think anyone worked on it. It will be good to add support for ARM64 chromium

marques-work:
@zabil I think Nivedha might be occupied ATM; do you by chance have any cycles to review this PR?

scripts/deps/update-package-json.js
```diff
@@ -0,0 +1,154 @@
...
+                   desirable as we use this block to pre-download transitive optionalDependencies to satisfy builds for linux
+                   agents in the build pipeline (currently, anyway), so upgrading these might break the build if the parent package
+                   is not using the latest version.
+
+`);
+
+  process.exit(1);
+}
+
+function update(dict) {
```
marques-work, line 88: This basically loops through a list of dependencies and asks the npm registry for a list of all versions.

For each package, use `semver` to determine the upgrade candidate using this logic:

- if `--canary` was specified, take the most recent version
- else, construct a `semver` constraint to find the latest version within that constraint

if the candidate is newer than that listed in `package.json`, replace the version value.

scripts/deps/update-package-json.js
```diff
@@ -0,0 +1,154 @@
...
+    const detectModifier = range.match(/^([^0-9]+)/);
+    const modifier = detectModifier ? detectModifier[1] : '';
+    const upgraded = modifier + latest;
+
+    if (range !== upgraded) {
+      dict[mod] = upgraded;
+      console.log('  => updating', mod, 'to', upgraded);
+    }
+  });
+}
```
marques-work, line 107: This basically loops through a list of dependencies and asks the npm registry for a list of all versions.

For each package, use `semver` to determine the upgrade candidate using this logic:

- if `--canary` was specified, take the most recent version
- else, construct a `semver` constraint to find the latest version within that constraint

if the candidate is newer than that listed in `package.json`, replace the version value.

scripts/updateChromium.js
```diff
@@ -1,9 +1,11 @@
 const https = require('https');
 const { writeFileSync, readFileSync } = require('fs');
 const { execSync } = require('child_process');
-const BrowserFetcher = require('./lib/browserFetcher');
+const BrowserFetcher = require(`${__dirname}/../lib/browserFetcher`);
 const supportedPlatforms = BrowserFetcher.supportedPlatforms;
 
+const PACKAGE_JSON = `${__dirname}/../package.json`;
```
marques-work, line 7: These changes just make the script idempotent from where and how it is invoked.

.editorconfig
```diff
@@ -0,0 +1,17 @@
+# top-most EditorConfig file
```
zabil, line 1: Do we need this in this PR?

scripts/deps/update-package-json.js
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env node
+
+/*
+ * Copyright (c) 2022 - Thoughtworks Inc. All rights reserved.
```
zabil, line 4: Can we drop this copyright?

scripts/updateChromium.js
zabil: Can we use path.join for this?

```suggestion
const PACKAGE_JSON = path.join(__dirname, "..", "package.json");
```

zabil (changes requested):
Added a few small comments. Will need to pull and run it locally meanwhile.

scripts/deps/update-package-json.js
marques-work: Sure, I copied this from my last project and overlooked this.

zabil (approved):

PR #2668, opened 2022-12-07
Add support for Mac ARM64 Chromium

marques-work:
@zabil quick review when you get a moment? I've tested this on my M1 MBP and `npm test` passes.

Also verified the binary image is indeed ARM64 (in case your machine is x86_64):

```console
$ file .local-chromium/mac_arm-1075999/chrome-mac/Chromium.app/Contents/MacOS/Chromium
.local-chromium/mac_arm-1075999/chrome-mac/Chromium.app/Contents/MacOS/Chromium: Mach-O 64-bit executable arm64
```


zabil (approved):
Thanks for this @marques-work

PR #2669, opened 2022-12-09
Bump version 1.3.6

marques-work:
Can we do a release to make the macOS arm64 enhancement available?

cc: @zabil
cc: @NivedhaSenthil

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "taiko",
-  "version": "1.3.5",
+  "version": "1.3.6",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "taiko",
-      "version": "1.3.5",
+      "version": "1.3.6",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


zabil:
@marques-work Yes! 

Will do it next week after seeing if someone pulls from master to give it a spin and report issues :)

marques-work:
Maybe wait for #2670 before bumping? `documentation` package just released an update that fixes a critcial vuln.

zabil (approved):

marques-work:
Thanks @zabil -- can we tag/cut a release so it's avail on npm? Or did you want to wait a little bit for any issues to come in?

zabil:
Oh dang, I was supposed to label that as a release candidate for it to automatically trigger the release. So need to see good I can trigger it manually

zabil:
@marques-work if you don't mind can you send another pr bumping up the version to 1.3.7?

marques-work:
I can -- do you think it would be better do revert to 1.3.5 and then bump up again so we don't skip a version number? Or is it just better to go to 1.3.7?

marques-work:
actually, 1.3.7 is just easier -- why go through 2 PRs right?

PR #2721, opened 2024-03-11
Update .eleventy.js

Alexlesch98:
In this refactored version:

1. I've grouped the required modules at the beginning for better clarity.
2. Each configuration or plugin addition is separated with comments for easier understanding.
3. The Markdown rendering configuration is now encapsulated in a variable (mdIt) for better modularity.
4. The conditional check in the eleventyConfig.addFilter('markdown', ...) function is simplified for readability.


zabil:
Closing and the dco is not signed

PR #2730, opened 2024-06-11
Upgrade gauge ts and chrome version

### LINKED ISSUE #2732: Browser processes do not terminate on Windows with Chrome(ium) version 126 and above
Using Chromium version 120.xxxx with Gauge and Taiko on Windows worked correctly. However, after upgrading to Chromium version 126.xxxx, it has been noticed that Chromium processes do not terminate after running tests on CI, leading to excessive resource consumption on the machine.

To address the issue, a stable build of 'Chrome for testing' was used, but the problem persisted.

Steps to Reproduce:
Install Chrome version 126 or higher.
Create and run a simple test several times.
Open Task Manager and note the active Chrome processes that remain.
Try to delete the node_modules folder to verify that the processes have indeed not ended (deletion will be blocked).

Expected Result:
Chrome processes should correctly terminate after test execution.

Actual Result:
Chrome processes remain hanging in the system, which blocks folder deletion and excessively consumes system resources.

Software Used:
Taiko: 1.3.10
Operating System: Windows 10
Node.js: v18.18.2

Link to download the used version of Chrome: 
https://storage.googleapis.com/chrome-for-testing-public/126.0.6478.126/win64/chromedriver-win64.zip

zabil:
- Bump gauge-ts plugin version
- Move windows builds to node 20
- Upgrade chrome version

Fixes #2732

### DIFF
--- a/lib/actions/pageActionChecks.js
+++ b/lib/actions/pageActionChecks.js
@@ -184,7 +184,7 @@ const waitAndGetActionableElement = async (selector, force, checks = defaultChec
       actionableElement = elements[0];
       return;
     }
-    if (elementsLength !== elements.length) {
+    if (elements && elementsLength !== elements.length) {
       throw Error('Found too many matches. Please use a selector that is more specific');
     }
     throw new Error(`${description(selector)}${error}`);
--- a/scripts/updateChromium.ts
+++ b/scripts/updateChromium.ts
@@ -1,6 +1,6 @@
-import { writeFileSync, readFileSync } from 'fs';
-import { execSync } from 'child_process';
-import path from 'path';
+import { execSync } from 'node:child_process';
+import { readFileSync, writeFileSync } from 'node:fs';
+import path from 'node:path';
 
 const PACKAGE_JSON_PATH = path.join(__dirname, '..', 'package.json');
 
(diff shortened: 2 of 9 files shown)


lib/actions/pageActionChecks.js
```diff
@@ -184,7 +184,7 @@ const waitAndGetActionableElement = async (selector, force, checks = defaultChec
       actionableElement = elements[0];
       return;
     }
-    if (elementsLength !== elements.length) {
+    if (elements && elementsLength !== elements.length) {
```
zabil, line 187: Adding this to avoid undefined error on `elements`

test/functional-tests/specs/data/IFrameElements.html
```diff
@@ -2,7 +2,7 @@
 <html>
 <body>
 
-<iframe src="./HTMLElements.html" width="1440" height="900">
+<iframe src="HTMLElements.html" width="800" height="600">
```
zabil, line 5: Use a reasonably sized viewport

scripts/updateChromium.ts
zabil, line 1: Fix typescript running issues on newer versions of node

NivedhaSenthil (approved):
Looks great !!

Issue #2735, opened 2024-07-19
Migrate from eslint to biome


zabil:
**Is your feature request related to a problem? Please describe.**
Taiko uses eslint in the current project. However many of the dependencies that eslint uses are
quickly being outdated and show issues in newer versions of node. For example

```
npm run test-docs                                                                                                                    15:08:34

> taiko@1.4.1 test-docs
> cd test/docs-tests && node prepare.js && eleventy && cd ./gauge && npm install && npm test

(node:13250) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
[11ty] Writing ./tmp/docs/_site/api/functionReturningPromiseOfObject/index.html from ./tmp/docs/api/api.njk
[11ty] Writing ./tmp/docs/_site/api/functionReturningPromise/index.html from ./tmp/docs/api/api.njk
```

```
npm list punycode                                                                                                                                                                              11:05:06
taiko@1.4.1 /Users/zabilcm/projects/taiko
└─┬ eslint@8.57.0
  └─┬ ajv@6.12.6
    └─┬ uri-js@4.4.1
      └── punycode@2.3.1
```

**Describe the solution you'd like**
Use [biome](https://biomejs.dev) instead of eslint. biome also bundles a formatter.

Note:

This would also involve fixing linting issues in the default setting of biome.

zabil:
Eslint is not used in the code anymore but is used in a testing library

```
npm list eslint
taiko@1.4.1 /Users/zabilcm/Gauge/taiko
└─┬ rewire@7.0.0
  └─┬ eslint@8.57.0
    └─┬ @eslint-community/eslint-utils@4.4.0
      └── eslint@8.57.0 deduped
```

There an open issue on rewire removing eslint. But I don't think that will happen anytime soon. https://github.com/jhnns/rewire/issues/156

Will need to look at alternatives to rewire to completely remove eslint dependency.

PR #2736, opened 2024-07-31
Biome migration

zabil:
Remove eslint and use biome for formatting a linting to reduce dependencies. Also fix the linting and formatting errors.

Refer #2735

### DIFF
--- a/.github/workflows/taiko.yml
+++ b/.github/workflows/taiko.yml
@@ -22,6 +22,10 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node_version }}
+      - name: Setup Biome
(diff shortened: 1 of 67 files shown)


zabil:
This gigantic PR is ready now. I hope it's ok if I merge this. It is mostly formatting and linting changes. I did some testing and there are no major issues. A few deprecation warning for old node api for example

```
> (node:6633) [DEP0048] DeprecationWarning: The `util.isError` API is deprecated. Please use `ObjectPrototypeToString(e) === "[object Error]" || e instanceof Error` instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

Which I feel can be fixed in another PR

PR #2747, opened 2024-12-20
Update click.js for relative selectors

DCoomer:
If a Relative Selector is passed in as the third argument inside of click(), it doesn't actually make it down to waitAndGetActionableElement. 

 We found that `await click(element, options, within(withinElement))`  doesn't search in the "within" element with the Biome Migration changes

Signed-off-by: DCoomer <deshuncoomer@gmail.com>

### DIFF
--- a/lib/actions/click.js
+++ b/lib/actions/click.js
@@ -49,11 +49,11 @@ async function simulateInputEvents(options) {
 }
 
 async function click(selector, options = {}, ...args) {
-  let allOptions = options;
+  let allOptions = args;
   if (options instanceof RelativeSearchElement) {
     allOptions = [options].concat(args);
   }
-  const clickOptions = setClickOptions(allOptions);
+  const clickOptions = setClickOptions(options);
   clickOptions.noOfClicks = clickOptions.clickCount || 1;
 
   if (isSelector(selector) || isString(selector) || isElement(selector)) {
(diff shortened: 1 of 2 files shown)


DCoomer:
Do I need to re-submit for the DCO?

lib/actions/click.js
```diff
@@ -49,11 +49,11 @@ async function simulateInputEvents(options) {
 }
 
 async function click(selector, options = {}, ...args) {
```
NivedhaSenthil, line 51: Can we please add a unit test for this case ? https://github.com/getgauge/taiko/blob/6645c6d884560cb1f2d33008e7ead49ff7d7d269/test/unit-tests/click.test.js#L149 a case with other click option and proximity selector

NivedhaSenthil:
Thanks for the PR :)

NivedhaSenthil (approved):

DCoomer:
Updated the formatting https://github.com/getgauge/taiko/pull/2747/commits/617b982bf32c2c1170d74512a097e16a7aa6004c

NivedhaSenthil:
Looks like it is still failing for format

DCoomer:
Updated Again

DCoomer:
Added the proper HTML for the unit test

PR #2764, opened 2026-01-28
Migrate to npm workspaces

zabil:
Restructured the project to use npm workspaces for better modularity
and library import capabilities. Updated build and lint workflows
to use Biome for improved performance and standardization. Upgraded
dependencies and resolved package conflicts. Enhanced testing pipeline
with optimized CI actions and proper dependency management.

Key changes:
- Migrated to npm workspaces architecture
- Implemented Biome for linting and formatting
- Upgraded all package dependencies
- Optimized CI/CD workflow configuration
- Resolved dependency conflicts and duplications
- Re-introduced documentation tests


.github/workflows/taiko.yml
```diff
@@ -40,10 +40,11 @@ jobs:
             libdrm-dev \
             libgbm-dev \
             libasound-dev
-      - name: unit-tests
-        run: |
-          npm run lint:ci
-          npm test
+      - name: Run Biome
+        uses: biomejs/setup-biome@v2
```
chadlwilson, line 44: FWIW, I'd changed this to use an npm lint task earlier to ensure the biome version matched that defined in the package.json so it could be run locally deterministically and to remove complexity/dependencies from the workflows/actions.

chadlwilson:
Apologies - thus was a bit big for me to get to, having only legacy experience with yarn workspaces and not npm ones.

Bit surprised it added so many lines though, I thought with workspaces the dependencies were consolidated at the root or something? Or maybe the GitHub changes summary is misleading.

zabil:
Hey @chadlwilson no worries. 

> Bit surprised it added so many lines though,

They can be consolidated for common packages and be specific for projects. 
Another big change with workspaces is the tests referring to the package instead of relative paths. For example 

before

```
require("../../lib/taiko")
```

after

```
require("taiko")
```

this means changing all tests

.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - name: Deploy on github
         run: |
-          version=$(grep '^  "version"' package.json | sed 's/"version":[ ]*"//' | sed 's/",//' | tr -d ' ')
+          version=$(grep '"version"' packages/taiko/package.json | sed 's/"version":[[:space:]]*"//' | sed 's/",//' | tr -d [:space:])
```
chadlwilson, line 34: Looks like you accidentally reverted my fix to the publishing here. This regex is wrong, and is causing the branch name/release name to concatenate with the chrome version. See https://github.com/getgauge/taiko/releases/tag/v1.4.7140.0.7339.82

I removed the character class usage since it was unnecessarily shell and `sed`-variant dependent making it annoying to sanity check the logic on MacOS.

```shell
$ grep '^  "version"' packages/taiko/package.json | sed 's/"version":[ ]*"//' | sed 's/",//' | tr -d ' '
1.4.8
```

netlify.toml
```diff
@@ -1,9 +1,7 @@
 [build]
-  command = "npm run doc"
-  publish = "docs/_site"
-[build.environment]
-  NODE_VERSION = "24"
```
chadlwilson, line 5: Looks like accidentally reverted fix during merge - netflify's default node is ancient and won't work (or at least not consistently with build).

.github/workflows/release.yml
zabil: Ah shoot, ok will look into these changes and raise a PR

chadlwilson: Yeah no big deal, I didn't realize you had a big PR in progress so sorry for the conflicts.

netlify.toml
zabil: I changed netlify node version via netlify console

> 4:23:04 PM: v22.22.0 is already installed.
> 4:23:04 PM: Now using node v22.22.0 (npm v10.9.4)

chadlwilson: Ok. Still feels nicer to drive it via code, but up to you 👍

chadlwilson:
Thx for your work on taiko. I've generally focused on "rest of gauge" since GoCD still uses it (via gauge-ruby) and only used taiko for a simple project in my earlier Thoughtworks life so it's good to have you keeping an eye on taiko 🙏

PR #2771, opened 2026-04-16
fix: handle empty quads from getContentQuads for proximity selectors  

### LINKED ISSUE #2757: Using any proximity selector in Taiko 1.4.3/1.4.4 throws TypeError: Cannot read properties of undefined (reading '1')
**Describe the bug**
Using any proximity selector in Taiko 1.4.3/1.4.4 throws TypeError: Cannot read properties of undefined (reading '1')

**To Reproduce**
Steps (or script) to reproduce the behavior:
I am using gauge 1.6.14 and Taiko 1.4.3. The same code mentioned below used to work well with GUAGE1.5.7 and TAIKO 1.0.28.
After upgrading, it is throwing the above error TypeError: Cannot read properties of undefined (reading '1')
Create a step with any of the below code, it throws the above error.
  
assert.ok(await text(content1, toRightOf(content2)).exists());

or

assert.ok(await text(content1, toLeftOf(content2)).exists());

or

 assert.ok(await text(content1, below(content2)).exists());

or

assert.ok(await text(content1, above(content2)).exists());

or 

assert.ok(await text(expectedText, near(neighbourText)).exists());

**Logs**
26-08-2025 16:02:47.937 [Gauge] [ERROR] Stacktrace: 
TypeError: Cannot read properties of undefined (reading '1')
    at Object.getBoundingClientRect (node_modules\taiko\lib\handlers\domHandler.js:152:28)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async RelativeSearchElement.condition (node_modules\taiko\lib\taiko.js:2029:20)
    at async RelativeSearchElement.validNodes (node_modules\taiko\lib\proximityElementSearch.js:65:11)
    at async getMatchingNode (node_modules\taiko\lib\proximityElementSearch.js:30:28)
    at async handleRelativeSearch (node_modules\taiko\lib\proximityElementSearch.js:13:25)
    at async TextWrapper._get (node_modules\taiko\lib\elementWrapper\textWrapper.js:24:14)
    at async node_modules\taiko\lib\elementSearch.js:358:22
    at async waitUntil (node_modules\taiko\lib\helper.js:135:11)
    at async node_modules\taiko\lib\elementSearch.js:356:7
```
Paste any log or error messages here
```

**Expected behavior**
All proximity selectors should work as expected.

**Screenshots**
Both the elements are getting highlighted as attached below, but not getting validated with the above code.

Image

**Versions:**
 - Taiko: 1.4.3
 - Gauge 1.6.14
 - OS Windows Server 2022 Standard
 - Node.js 22.16.0

```
gauge -v
Gauge version: 1.6.14
Commit Hash: 9561cf7

Plugins
-------
html-report (4.3.3)
java (0.8.1)
js (5.0.1)
screenshot (0.3.2)
```

**Additional context**

winst0niuss:
Fixes #2757                                    
                                                                                                                                                                                                                                                                                                                              
  ## Problem                                                                           
                                               
  Proximity selectors (`toRightOf`, `toLeftOf`, `above`, `below`, `near`) throw `TypeError: Cannot read properties of undefined (reading '1')` in `domHandler.js:152`.                                                                                                                                                        
   
  ### Root cause                                                                                                                                                                                                                                                                                                              
                                                                                       
  After migrating from `dom.getBoxModel` to `dom.getContentQuads`, the null-check logic broke. When `getContentQuads` returns `quads: []` (element is hidden, off-screen, etc.), `result.quads[0]` is `undefined`, but `getBoxModel` still returned a non-null object — so the `if (!result)` guard in `getBoundingClientRect`
   never fired, and accessing `quad[1]` crashed.                                       
                                                                                                                                                                                                                                                                                                                              
  Two additional call sites had the same unguarded pattern:                                                                                                                                                                                                                                                                   
  - `getPositionalDifference` — did not handle `null` from `getBoundingClientRect`
  - `highlightElement` — did not handle `null` from `getBoxModel`                                                                                                                                                                                                                                                             
                                                                                       
  ## Changes                                                                                                                                                                                                                                                                                                                  
                                                                                       
  **`domHandler.js` — `getBoxModel`**: return `null` when `quads` is empty, making all downstream guards work correctly.                                                                                                                                                                                                      
                                         
  **`domHandler.js` — `getPositionalDifference`**: return `Number.POSITIVE_INFINITY` when either element has no bounding box, so it never wins the proximity comparison in `proximityElementSearch.js`.                                                                                                                       
                                                                                       
  **`elementHelper.js` — `highlightElement`**: skip highlight gracefully when `getBoxModel` returns `null` (e.g. visible element with no rendered quads on certain platforms).                                                                                                                                                
                                                                                       
  **`tests/unit-tests/proximitySelector.test.js`**: new regression test covering all five proximity selectors and a hidden element case (`display:none`) that triggers the empty-quads path.

### DIFF
--- a/packages/taiko/lib/elements/elementHelper.js
+++ b/packages/taiko/lib/elements/elementHelper.js
@@ -16,6 +16,11 @@ const highlightElement = async (element) => {
       await overlayHandler.highlightQuad(result.model.border);
       await wait(1000);
       await overlayHandler.hideHighlight();
+    } else {
+      console.warn(
+        "WARNING: Taiko cannot highlight element — no rendered content quads " +
+          "(element may be off-screen, zero-sized, or not yet laid out).",
+      );
     }
   } else {
     console.warn("WARNING: Taiko cannot highlight hidden elements.");
--- a/packages/taiko/lib/elements/elementHelper.js
+++ b/packages/taiko/lib/elements/elementHelper.js
@@ -12,9 +12,11 @@ const highlightElement = async (element) => {
 
   if (await element.isVisible()) {
     const result = await domHandler.getBoxModel(element.get());
-    await overlayHandler.highlightQuad(result.model.border);
-    await wait(1000);
-    await overlayHandler.hideHighlight();
+    if (result) {
+      await overlayHandler.highlightQuad(result.model.border);
+      await wait(1000);
+      await overlayHandler.hideHighlight();
+    }
   } else {
     console.warn("WARNING: Taiko cannot highlight hidden elements.");
   }
--- a/tests/unit-tests/proximitySelectors/proximitySelector.test.js
+++ b/tests/unit-tests/proximitySelectors/proximitySelector.test.js
@@ -19,7 +19,7 @@ const {
   removeFile,
   openBrowserArgs,
   resetConfig,
-} = require("./test-util");
+} = require("../test-util");
 const test_name = "proximitySelector";
 
 describe(test_name, () => {
--- a/.github/workflows/taiko.yml
+++ b/.github/workflows/taiko.yml
@@ -45,7 +45,16 @@ jobs:
         uses: biomejs/setup-biome@v2
       - run: biome check packages/taiko
       - name: Run unit tests
+        if: matrix.os != 'windows-latest'
         run: npm run test:unit:silent
+
+      - name: Run unit tests (windows, with retry)
+        if: matrix.os == 'windows-latest'
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 10
+          max_attempts: 2
+          command: npm run test:unit:silent
 
   functional-tests:
     needs: unit-tests
@@ -82,7 +91,16 @@ jobs:
             libgbm-dev \
             libasound-dev
       - name: functional-tests
+        if: matrix.os != 'windows-latest'
         run: npm run test-functional
+
+      - name: functional-tests (windows, with retry)
+        if: matrix.os == 'windows-latest'
+        uses: nick-fields/retry@v3
+        with:
+          timeout_minutes: 15
+          max_attempts: 2
+          command: npm run test-functional
 
       - name: Upload html report
         uses: actions/upload-artifact@v6
--- a/packages/taiko/lib/handlers/domHandler.js
+++ b/packages/taiko/lib/handlers/domHandler.js
@@ -139,6 +139,9 @@ async function getBoxModel(e) {
   const result = await dom.getContentQuads({
     objectId: isElement(e) ? e.objectId : e,
   });
+  if (!result.quads || result.quads.length === 0) {
+    return null;
+  }
   return { model: { border: result.quads[0] } };
 }
 
@@ -159,6 +162,9 @@ async function getBoundingClientRect(e) {
 const getPositionalDifference = async (nodeA, nodeB) => {
   const r = await getBoundingClientRect(nodeA);
   const v = await getBoundingClientRect(nodeB);
+  if (!r || !v) {
+    return Number.POSITIVE_INFINITY;
+  }
   const topDiff = Math.abs(r.top - v.top);
   const leftDiff = Math.abs(r.left - v.left);
   const bottomDiff = Math.abs(r.bottom - v.bottom);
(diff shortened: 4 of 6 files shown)


packages/taiko/lib/elements/elementHelper.js
```diff
@@ -12,9 +12,11 @@ const highlightElement = async (element) => {
 
   if (await element.isVisible()) {
     const result = await domHandler.getBoxModel(element.get());
-    await overlayHandler.highlightQuad(result.model.border);
-    await wait(1000);
-    await overlayHandler.hideHighlight();
+    if (result) {
```
NivedhaSenthil, line 15: Do we know why the bounding box is not available even though the element is marked visible ?

NivedhaSenthil: Knowing the root cause can help warning the user for a potential fix

packages/taiko/lib/elements/elementHelper.js
```diff
@@ -12,9 +12,11 @@ const highlightElement = async (element) => {
...
+      await overlayHandler.highlightQuad(result.model.border);
+      await wait(1000);
+      await overlayHandler.hideHighlight();
+    }
   } else {
     console.warn("WARNING: Taiko cannot highlight hidden elements.");
```
NivedhaSenthil, line 26: If the result is not available because the element is hidden, we will have to warn using this case

tests/unit-tests/proximitySelectors/proximitySelector.test.js
```diff
@@ -0,0 +1,99 @@
+const chai = require("chai");
```
NivedhaSenthil, line 1: Please move the tests inside `tests/unit-tests/proximitySelectors` and if possible preferably to respective proximity selector test files.

packages/taiko/lib/elements/elementHelper.js
winst0niuss: element.isVisible() checks DOM/CSS properties (display, visibility, opacity, dimensions), while DOM.getContentQuads relies on the render tree and returns an empty array when the element has no painted area — e.g. it's off-screen, zero-sized, or under transform: scale(0). So visible === true && quads.length === 0 is a valid Chrome CDP state, not a Taiko bug.
 
I've added a console.warn for this case so the user gets a hint instead of silent skip.

packages/taiko/lib/elements/elementHelper.js
winst0niuss: Added a separate console.warn inside the if (!result) branch:
"WARNING: Taiko cannot highlight element — no rendered content quads (element may be off-screen, zero-sized, or not yet laid out)."
This is distinct from the existing hidden-element warning.

NivedhaSenthil (approved):

tests/unit-tests/proximitySelectors/proximitySelector.test.js
winst0niuss: Moved the test file to tests/unit-tests/proximitySelectors/proximitySelector.test.js.

winst0niuss:
@NivedhaSenthil , check pls

winst0niuss:
After merging getgauge:master, the Windows CI jobs started failing due to a bug introduced in #2772.           
 
**Root cause:** evaluate.test.js (added in #2772) uses .rejectedWith() from chai-as-promised without importing it. On macOS/Linux the tests pass by accident — $.test.js runs first (glob sorts $ before letters in ASCII) and loads chai-as-promised globally. On Windows, glob returns files in a different order, so evaluate.test.js runs before chai-as-promised is loaded, causing TypeError: rejectedWith is not a function.

winst0niuss:
**Fix:** Added an explicit chai-as-promised import to evaluate.test.js so the test no longer depends on execution order.

winst0niuss:
Added a retry (2 attempts) for Windows CI jobs to avoid flaky failures we keep seeing on GitHub runners
Other jobs unchanged

winst0niuss:
@zabil , re-run tests pls

winst0niuss:
All checks are green. Please merge 🙂

zabil:
@NivedhaSenthil please merge if you are ok with the review updates.

NivedhaSenthil (approved):

PR #2772, opened 2026-04-17
Fix: evaluate returns empty object when function throws (#2613)

### LINKED ISSUE #2613: evaluate returns empty object when function throws
**Describe the bug**

```js
const x = await evaluate(() => {
  return document.querySelector('.doesNotExist').textContent
})
```

When no element with class `doesNotExist` exists `x` will have value `{}`. I think this is because it's a json-ified error. 

**To Reproduce**

```js
const { openBrowser, evaluate, goto } = require('taiko');
openBrowser();
goto('https://example.com')
const x = await evaluate(() => {
  return document.querySelector('.doesNotExist').textContent
})
console.log(x);
```

**Expected behavior**

`evaluate` should throw when the client function throws. It might require wrapping the function and returning an either type which includes the message and the stack of the error.

NajoroRabiaza:
Description
This PR fixes the issue where evaluate swallows exceptions thrown inside the browser context and returns an empty object.

Changes:

Checked for result.exceptionDetails in CDP response inside doActionAwaitingNavigation.
If an exception occurred, throw a standard Error in the Node context with the browser's serialized error message.
Also added a unit test to prevent future regressions.

Fixes #2613

### DIFF
--- a/packages/taiko/lib/taiko.js
+++ b/packages/taiko/lib/taiko.js
@@ -2367,6 +2367,15 @@ module.exports.evaluate = async (selector, callback, options = {}) => {
     });
   });
 
+  if (result && result.exceptionDetails) {
+    const exceptionDetails = result.exceptionDetails;
+    const errorMessage =
+      exceptionDetails.exception && exceptionDetails.exception.description
+        ? exceptionDetails.exception.description
+        : exceptionDetails.text || "Error in evaluate callback";
+    throw new Error(errorMessage);
+  }
+
   if (!_options.silent) {
     descEvent.emit("success", "Evaluated given script");
   }
--- a/packages/taiko/lib/taiko.js
+++ b/packages/taiko/lib/taiko.js
@@ -2370,9 +2370,9 @@ module.exports.evaluate = async (selector, callback, options = {}) => {
   if (result && result.exceptionDetails) {
     const exceptionDetails = result.exceptionDetails;
     const errorMessage =
-      exceptionDetails.exception && exceptionDetails.exception.description
-        ? exceptionDetails.exception.description
-        : exceptionDetails.text || "Error in evaluate callback";
+      exceptionDetails.exception?.description ??
+      exceptionDetails.text ??
+      "Error in evaluate callback";
     throw new Error(errorMessage);
   }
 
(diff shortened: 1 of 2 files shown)


packages/taiko/lib/taiko.js
```diff
@@ -2367,6 +2367,15 @@ module.exports.evaluate = async (selector, callback, options = {}) => {
     });
   });
 
+  if (result && result.exceptionDetails) {
+    const exceptionDetails = result.exceptionDetails;
+    const errorMessage =
+      exceptionDetails.exception && exceptionDetails.exception.description
+        ? exceptionDetails.exception.description
+        : exceptionDetails.text || "Error in evaluate callback";
```
zabil: Can we change this to
```
 const errorMessage =
   exceptionDetails.exception?.description ??
   exceptionDetails.text ??
   "Error in evaluate callback";
```

tests/unit-tests/evaluate.test.js
```diff
@@ -103,6 +103,19 @@ describe(testName, () => {
       expect(actual).to.equal(testName);
     });
 
+    it("should throw when the callback throws an error", async () => {
```
zabil, line 106: Have you tried 

```
 it("should throw when the callback throws an error", async () => {
   await expect(
     evaluate(() => document.querySelector(".doesNotExist").textContent)
   ).to.be.rejectedWith(Error, /TypeError/);
 });
```

zabil (changes requested):

packages/taiko/lib/taiko.js
NajoroRabiaza: Yes, that's updated

tests/unit-tests/evaluate.test.js
NajoroRabiaza: Yeah tried it, Didn't think of rejectedWith, cleaner indeed. Updated!

zabil:
@NajoroRabiaza can you please sign the dco? https://github.com/getgauge/taiko/pull/2772/checks?check_run_id=71931483978

NivedhaSenthil (approved):

zabil (approved):

PR #2773, opened 2026-04-17
Fix: wait for element to be stable fails on infinite animations (#2711)

### LINKED ISSUE #2711: Click on Moving links is not waiting for the link to stop moving
Description

On HTML page with a link that is constantly moving, the click function doesn't wait for the link to stop moving

Steps to reproduce
openBrowser()
goto(html)
click("Moving Link")

expected:
to wait 10 seconds as indicated here:https://github.com/getgauge/taiko/blob/master/lib/actions/pageActionChecks.js#L71-L107

Actual:
clicks right away after a second or so

reproduced every time, on lots of speeds 0.1s, 1s and for 500 px 

The html:
```
<!DOCTYPE html>
<html>
<head>
  <title>Moving Link Example</title>
  <style>
    /* Define the animation */
    @keyframes moveLink {
      0% { transform: translateX(0); }
      100% { transform: translateX(200px); }
    }

    /* Apply the animation to the link */
    a {
      display: inline-block;
      animation: moveLink 2s linear infinite; /* The link will move every 2 seconds */
    }
  </style>
</head>
<body>
  <a href="#">Moving Link</a>
</body>
</html>
```

seems very similar to this - https://github.com/getgauge/taiko/issues/2550

NajoroRabiaza:
Fixes #2711

This PR fixes an issue where actions like click() on elements with infinite CSS animations would not correctly wait for the 10-second timeout and would fail prematurely or proceed unpredictably.

Root Cause
The root cause was a combination of logic flaws in how element stability was checked and how timeouts were communicated back through the Chrome DevTools Protocol (CDP):

isInSamePosition was returning an integer instead of a boolean, confusing the stability logic.

A missing return statement after the initial requestAnimationFrame caused the comparison to happen immediately against the same object.

The 10-second timeout rejected with a primitive false instead of an Error object. In CDP, rejecting with a primitive value swallows the error (resulting in undefined), meaning the Node process never saw the timeout exception.

checkStable was not checking for res.exceptionDetails to propagate the error properly.

Changes Made
Updated isInSamePosition to return a strict boolean (topDiff + leftDiff + bottomDiff + rightDiff === 0).

Added the missing return; statement in waitForElementToBeStable after the first requestAnimationFrame call.

Modified the 10-second timeout to reject with a proper new Error("Element is not stable: still moving after 10000ms").

Added an exceptionDetails check inside checkStable to correctly intercept the timeout error from the browser context and throw it as a standard JavaScript Error in Node.js.

### DIFF
--- a/packages/taiko/lib/actions/pageActionChecks.js
+++ b/packages/taiko/lib/actions/pageActionChecks.js
@@ -79,7 +79,10 @@ const checkStable = async (elem) => {
   function waitForElementToBeStable() {
     let elem = this;
     return new Promise((resolve, reject) => {
-      setTimeout(() => reject(false), 10000);
+      setTimeout(
+        () => reject(new Error("Element is not stable: still moving after 10000ms")),
+        10000,
+      );
       let previousRect;
       let currentRect;
       function isInSamePosition(previousRect, currentRect) {
@@ -87,7 +90,7 @@ const checkStable = async (elem) => {
         const leftDiff = Math.abs(previousRect.left - currentRect.left);
         const bottomDiff = Math.abs(previousRect.bottom - currentRect.bottom);
         const rightDiff = Math.abs(previousRect.right - currentRect.right);
-        return topDiff + leftDiff + bottomDiff + rightDiff;
+        return topDiff + leftDiff + bottomDiff + rightDiff === 0;
       }
       (function step() {
         if (elem.nodeType === Node.TEXT_NODE) {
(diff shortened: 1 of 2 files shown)


packages/taiko/lib/actions/pageActionChecks.js
```diff
@@ -79,15 +79,18 @@ const checkStable = async (elem) => {
   function waitForElementToBeStable() {
     let elem = this;
     return new Promise((resolve, reject) => {
-      setTimeout(() => reject(false), 10000);
+      setTimeout(
```
NivedhaSenthil, line 82: Can you please add an unit test covering this case ?

NivedhaSenthil (changes requested):

packages/taiko/lib/actions/pageActionChecks.js
NajoroRabiaza: Yes, I have add two unit tests, so one for when the element is stable and one for the infinite animation case

NivedhaSenthil (approved):

NivedhaSenthil:
can you please fix the formatting issues ?

NajoroRabiaza:
Sure, it's done and I've already fixed the formatting

PR #2775, opened 2026-04-25
debug: capture taiko:intercept logs on Windows to diagnose navigation hang

zabil:
## Problem

The Windows CI job `FTs - NodeJS 22 & windows-latest` is consistently failing at `intercept.spec:38`:

```
Failed Step: Navigate to "https://localhost/employees/2/address"
Error Message: Navigation took more than 60000ms. Please increase the navigationTimeout.
```

Both retry attempts fail, so this is not random flakiness. The root cause is unknown — we don't know whether:

- **Scenario A**: `Fetch.requestPaused` fires but `fulfillRequest` fails silently (request stays paused → navigation hangs)
- **Scenario B**: `Fetch.requestPaused` never fires at all (Chrome bypasses the intercept for HTTPS on Windows)

### `packages/taiko/lib/logger.js`
Adds a `taiko:intercept` debug namespace (uses the existing `debug` package — zero overhead when `DEBUG` is unset).

### `packages/taiko/lib/handlers/fetchHandler.js`
Adds `logIntercept` calls at every key point in the intercept lifecycle:
- When an interceptor is registered (`addInterceptor`)
- When `requestPaused` fires (with the request URL)
- When no matching interceptor is found
- When a matching interceptor is found (with type)
- When `fulfillRequest`/`continueRequest` is called (with response details)
- When those CDP calls fail

### `.github/workflows/taiko.yml`
- Enables `DEBUG=taiko:intercept` on the Windows functional-test step so the logs appear inline in the CI job output on failure
- **Removes** the `nick-fields/retry` workaround from both unit-tests and functional-tests on Windows — retries were hiding the failure signal

## How to read the output

After this PR, the failing CI run will show one of:

**Scenario A** (requestPaused fires, fulfillRequest fails):
```
taiko:intercept requestPaused url=https://localhost/employees/2/address
taiko:intercept matched interceptor url=... action=object
taiko:intercept calling fulfillRequest (object) for url=...
taiko:intercept fulfillRequest/continueRequest failed for url=...
```

**Scenario B** (requestPaused never fires):
```
taiko:intercept requestPaused url=https://localhost/employees/1/address   ← appears
taiko:intercept requestPaused url=https://localhost/employees/1/address   ← appears (second nav)
                                                                          ← nothing for /2/address
```

This PR is diagnostic only — no behaviour change.

### DIFF
--- a/.github/workflows/taiko.yml
+++ b/.github/workflows/taiko.yml
@@ -48,13 +48,9 @@ jobs:
         if: matrix.os != 'windows-latest'
         run: npm run test:unit:silent
 
-      - name: Run unit tests (windows, with retry)
+      - name: Run unit tests (windows)
         if: matrix.os == 'windows-latest'
-        uses: nick-fields/retry@v3
-        with:
-          timeout_minutes: 10
-          max_attempts: 2
-          command: npm run test:unit:silent
+        run: npm run test:unit:silent
 
   functional-tests:
     needs: unit-tests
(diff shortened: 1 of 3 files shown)


zabil:
Fixed via #2776

Issue #2687, opened 2023-02-07
Need to pass only args to evaluate callback


combinesoma:
Hi Team - In the current version of the evaluate API, there is no provision to send only args to the callback, we need to pass an element along with the args param, is there any way where we can pass only arguments to the callback function?

Example: 
await evaluate( (args) => { }, options)

mzissa:
I agree with you. I faced this issue as well. The way I got around it is by passing a dummy select like this:

```
const dummyDiv = $("div");
await evaluate(dummyDiv, function (x) { });
```

PR #2777, opened 2026-04-27
fix: allow evaluate callback to receive args without selector (#2687)

NajoroRabiaza:
The issue was that when calling evaluate without a selector, there was no clean way to pass args directly to the callback. 

You had to write (element, args) => {} even if you didn't need the element at all.

The fix adds a noSelector flag that is set to true when no selector is passed. Inside evalFunc which runs in the browser context, if noSelector is true, the callback is called with just args as the first argument instead of (element, args). 

This way you can now write:
```js
await evaluate((args) => { return args[0] + args[1]; }, { args: [10, 20] })
```
The existing behavior with a selector or without args is not changed at all.

### DIFF
--- a/packages/taiko/lib/taiko.js
+++ b/packages/taiko/lib/taiko.js
@@ -2439,9 +2439,11 @@ module.exports.evaluate = async (selector, callback, options = {}) => {
   let elem;
   let _options = options;
   let _callback = callback;
+  let _noSelector = false;
   if (isFunction(selector)) {
     _options = callback || options;
     _callback = selector;
+    _noSelector = true;
     elem = (await $$xpath("//*"))[0];
   } else {
     elem = await findFirstElement(selector);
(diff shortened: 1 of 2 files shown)


zabil:
@NajoroRabiaza can you give a better use case on how this change is going to be used?

Also note this change will break the typescript definition of this function. At this point we do not want to change api's.

NajoroRabiaza:
Thanks for the feedback! I understand about the TypeScript definition concern. And a real use case would be something like injecting configuration or environment data into the page without needing a specific element, for example await evaluate((args) => { window.__config = args; }, { args: [{ env: 'test' }] }). But I understand if this is not the time to change the API.

NajoroRabiaza:
So we'll close this Pull Request ?

zabil:
Yes, sorry about that. For the sake of backward compatibility and a stable api. I would recommend closing this.

PR #2778, opened 2026-05-15
Handle text nodes in writable checks

mzl2233:
## Summary
- Resolve writable checks against the parent element when a text selector resolves to a text node
- Add unit coverage for contenteditable text-node writability

## Tests
- npx mocha tests/unit-tests/elements/element.test.js


zabil:
@mzl2233 can you please fix the formatting and sign the [DCO](https://github.com/getgauge/taiko/pull/2778/checks?check_run_id=76157262893)?

PR #2779, opened 2026-05-25
Keep TLS verification enabled for proxied browser downloads

darkdoomrag3:
## Summary

  This PR keeps TLS certificate validation enabled when Taiko downloads bundled Chromium through a configured proxy.

 Previously, `packages/taiko/lib/browser/fetcher.js` set `rejectUnauthorized = false` whenever `proxy-from-env` returned a proxy. That disabled certificate validation during
  install-time Chromium downloads.

  ## Security impact

  A user installing Taiko behind a configured proxy could receive and extract a forged Chromium archive without TLS certificate verification failing. Because the downloaded
  archive is later used as the browser executable, this creates a supply-chain code execution risk in environments where proxy settings can be influenced or where an untrusted
  proxy is used.

  ## Changes

  - Removed `options.rejectUnauthorized = false` from proxied Chromium download requests.
  - Kept proxy support through `https-proxy-agent`.
  - Added a regression test that asserts proxied Chromium download requests keep TLS verification enabled.

  ## Verification

  - Added unit test coverage in `tests/unit-tests/browserFetcher.test.js`.
  - Verified the focused regression test passes locally.

  ## Follow-up hardening

  A separate follow-up could add SHA-256 checksum verification for downloaded Chromium archives before extraction.


SECURITY_REVIEW_REPORT.md
zabil, line 1: @darkdoomrag3 can you please remove this file and add these details as a part of the PR description?

darkdoomrag3: Done, removed `SECURITY_REVIEW_REPORT.md` and added the details to the PR description.

zabil: I don't seem the description updated yet can you please check it again?
Also I ran the build and there are failures that might be related to this change

